PR #2, opened 2016-05-29
Display current email on control/editemailpassword

taedixon:
This is so we can tell people to check if their email address is up to date and if not they can date it.

### DIFF
--- a/templates/control/edit_emailpassword.html
+++ b/templates/control/edit_emailpassword.html
@@ -6,7 +6,7 @@ $:{TITLE("Edit Password and Email Address", "Settings", "/control")}
   <form name="controleditemailpassword" class="form skinny clear" action="/control/editemailpassword" method="post">
     $:{CSRF()}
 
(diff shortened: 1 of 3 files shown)


templates/control/edit_emailpassword.html
```diff
@@ -1,10 +1,14 @@
+$def with (email)
 $:{TITLE("Edit Password and Email Address", "Settings", "/control")}
 <div class="content">
   <p style="padding-top:2em;font-size:1.2em;">On this page you can change your associated email address, password, or both; to change only one of the items, leave the other pair of fields blank. If you are changing your password, please remember to follow the security guidelines described below.</p>
 
   <form name="controleditemailpassword" class="form skinny clear" action="/control/editemailpassword" method="post">
     $:{CSRF()}
 
+    <label for="currentemail">Current Email Address</label>
```
charmander: `<label>`s only apply to form input elements; could you make a `.label` class that’s styled the same way?

PR #3, opened 2016-05-30
Remove some unused parts

charmander:
Functions in `weasyl.files` and use of the `loginaddress` table, which is empty in production.


strain-113:
+1

PR #15, opened 2016-05-30
Accessible oauth

taedixon:
Changes to allow Weasyl users to generate their own OAuth2 client ID's and client secrets  (with limited scopes). Relies on Weasyl/libweasyl#3
Also modifies  the OAuth clickthrough page to display more information about the various scopes the application is requesting.

The registration and management UI is a bit (read: a lot) brutish so feel free to tear it apart but it should be at least basically functional. it's a developer thing anyway so getting it _beautiful_ is not my highest priority.

### DIFF
--- a/templates/control/edit_apikeys.html
+++ b/templates/control/edit_apikeys.html
@@ -103,6 +103,9 @@ $:{RENDER("common/stage_title.html", ["API Keys", "Settings"])}
       <label for="client-name">Application Name</label>
       <input id="client-name" type="text" name="client-name" class="input" required />
 
+      <label for="client-homepage">Application Homepage</label>
+      <input id="client-homepage" type="url" name="client-homepage" class="input" placeholder="optional" />
+
       <label for="redirect-uris">Redirect URIs</label>
       <input id="redirect-uris" type="url" name="redirect-uris" class="input" required
           placeholder="separate multiple URIs with a space"/>
@@ -125,7 +128,10 @@ $:{RENDER("common/stage_title.html", ["API Keys", "Settings"])}
         $for app in my_applications:
           <label class="input-checkbox">
             <input type="checkbox" name="clientid" value="${app.clientid}" />
-            <strong>${app.description}</strong>
+            $if app.homepage:
+              <a href="${app.homepage}"><strong>${app.description}</strong></a>
+            $else:
+              <strong>${app.description}</strong>
             <p>client ID: <tt>${app.clientid}</tt></p>
             <p>client secret: <tt>${app.client_secret}</tt></p>
             <p>scopes: <tt>${", ".join(app.scopes)}</tt></p>
--- a/templates/control/edit_apikeys.html
+++ b/templates/control/edit_apikeys.html
@@ -21,7 +21,7 @@ $:{RENDER("common/stage_title.html", ["API Keys", "Settings"])}
         However, user agents accessing Weasyl with your API key also
         will <em>never</em> be able to see your password.
       </p>
-      <button name="add-api-key" value="y" class="button last-input positive" style="float: right;">
+      <button name="add-api-key" value="y" class="button positive" style="float: right;">
         New API Key
       </button>
     </form>
@@ -37,7 +37,7 @@ $:{RENDER("common/stage_title.html", ["API Keys", "Settings"])}
           $for token, description in api_keys:
             <li>
               <label class="input-checkbox">
-                <input type="checkbox" name="delete-api-keys" value="${token}" /></th>
+                <input type="checkbox" name="delete-api-keys" value="${token}" />
 
                 <em>
                   $if description:
@@ -104,14 +104,26 @@ $:{RENDER("common/stage_title.html", ["API Keys", "Settings"])}
       <input id="redirect-uris" type="text" name="redirect-uris" class="input" required />
 
       <button name="create-oauth2-client" value="y" style="float:right;margin-top:1em"
-              class="button last-input positive">Register Application</button>
+              class="button positive">Register Application</button>
     </form>
 
     <h3>Manage OAuth2 Applications</h3>
       $if my_applications:
-        <form>
+        <form action="" method="POST" class="clear">
+        $:{CSRF()}
         $for app in my_applications:
-          ${app.description}<br>
+          <label class="input-checkbox">
+            <input type="checkbox" name="clientid" value="${app.clientid}" />
+            <strong>${app.description}</strong>
+            <p>client ID: <tt>${app.clientid}</tt></p>
+            <p>client secret: <tt>${app.client_secret}</tt></p>
+            <p>scopes: <tt>${", ".join(app.scopes)}</tt></p>
+          </label>
+
+        <button name="remove-oauth2-client" value="y" style="float:right;margin-top:1em"
+                class="button negative">Delete Selected Applications</button>
+        <button name="reissue-client-secret" value="y" style="float:right;margin-top:1em"
+                class="button">Reissue Secrets</button>
         </form>
       $else:
         <p style="padding-top: 1em;">You have no registered applications!</p>
--- a/templates/control/edit_apikeys.html
+++ b/templates/control/edit_apikeys.html
@@ -56,6 +56,7 @@ $:{RENDER("common/stage_title.html", ["API Keys", "Settings"])}
         <button class="button negative" style="float: right;">Delete Selected Keys</button>
       $else:
         <p style="padding-top: 1em;">You have no API keys!</p>
+
     </form>
 
 
--- a/templates/control/edit_apikeys.html
+++ b/templates/control/edit_apikeys.html
@@ -1,4 +1,4 @@
-$def with (api_keys, consumers)
+$def with (api_keys, consumers, oauth_scopes, my_applications)
 $:{RENDER("common/stage_title.html", ["API Keys", "Settings"])}
 
 <div id="apikeys-content" class="content form clear">
@@ -25,9 +25,8 @@ $:{RENDER("common/stage_title.html", ["API Keys", "Settings"])}
         New API Key
       </button>
     </form>
-  </div>
 
-  <div class="column">
+
     <h3>Manage API Keys</h3>
 
     <form name="manage-api-keys" action="" method="post" class="clear">
@@ -57,9 +56,11 @@ $:{RENDER("common/stage_title.html", ["API Keys", "Settings"])}
         <button class="button negative" style="float: right;">Delete Selected Keys</button>
       $else:
         <p style="padding-top: 1em;">You have no API keys!</p>
-
     </form>
 
+  </div>
+
+  <div class="column">
     <h3>Manage OAuth2 Consumers</h3>
 
     <form action="" method="post" class="clear">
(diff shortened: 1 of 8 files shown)


templates/control/edit_apikeys.html
```diff
@@ -78,16 +88,65 @@
...
+    </form>
 
+    <h3>Register OAuth2 Application</h3>
+    <form action="" method="POST" class="clear">
+      $:{CSRF()}
+      <p style="padding-top: 1em;">
+        If you are a developer, you may register an application as an OAuth2 consumer.
+        You can then use your client id and client secret to request authorization tokens from
+        Weasyl. Your client secret MUST be stored securely and should not be accessible
+        by your application's users under any circumstance. For more information see
```
strain-113: Add: Anyone with your secret effectively has control over your account. Do not share it or check it into source control, etc.

templates/control/edit_apikeys.html
```diff
@@ -78,16 +88,65 @@
...
-  </p>
+    <h3>Manage OAuth2 Applications</h3>
+      $if my_applications:
+        <form action="" method="POST" class="clear">
+        $:{CSRF()}
+        $for app in my_applications:
+          <label class="input-checkbox">
+            <input type="checkbox" name="clientid" value="${app.clientid}" />
+            $if app.homepage:
+              <a href="${app.homepage}"><strong>${app.description}</strong></a>
```
strain-113: Either inline or via CSS, this should probably be `target="_blank"

templates/oauth2/authorize.html
```diff
@@ -213,13 +220,22 @@
...
+      should be authorized to make requests on your behalf with the following permissions:</p>
+
+    $if scopes:
+      <table id="permissions">
+        <tr><td><strong>PERMISSION</strong></td><td><strong>GRANTS</strong></td></tr>
+        $for scope in scopes:
+          <tr><td>${scope['name']}</td><td>${scope['description']}</td></tr>
+      </table>
+    $else:
+      <p>No elevated permissions requested</p>
```
strain-113: This is ambiguous.

templates/oauth2/authorize.html
```diff
@@ -213,13 +220,22 @@
...
 
-    <p>Note that this means that <strong>${client.description}</strong> can
-      perform <em>almost any action</em> as if you were logged in. Do
-      not authorize this application if you do not trust it. However, this
+    <p>
+      Do not authorize this application if you do not trust it. However, this
       application will <em>never</em> be able to see or change your
```
strain-113: Why is this second sentence necessary? I'd rather people err on the side of paranoia.

weasyl/controllers/settings.py
```diff
@@ -1,5 +1,5 @@
 import os
-
+import re
```
strain-113: Is this necessary?

```
Python 2.7.10 (default, Oct 14 2015, 16:09:02) 
[GCC 5.2.1 20151010] on linux2
>>> whitespace_string = "This is  a\twhitespace\n\t\n\tseparated\r\nstring."
>>> whitespace_string
'This is  a\twhitespace\n\t\n\tseparated\r\nstring.'
>>> print whitespace_string
This is  a	whitespace
	
	separated
string.
>>> whitespace_string.split()
['This', 'is', 'a', 'whitespace', 'separated', 'string.']
```

weasyl/controllers/api.py
```diff
@@ -457,6 +462,7 @@ def GET(self):
 
 class api_favorite_(api_base):
     login_required = True
+    oauth_scopes = ['favourite']
```
strain-113: I admire commonwealth English, but 'favorite' is going to be the less surprising spelling so we should use that. In particular, it's the spelling used throughout the site's interface.

weasyl/define.py
```diff
@@ -343,7 +343,7 @@ def captcha_verify(form):
     return result[0] == 'true'
 
 
-def get_userid(sessionid=None):
+def get_userid(scopes=None):
     """
     Returns the userid corresponding to the user's sessionid; if no such session
```
strain-113: This docstring needs an update now.

weasyl/oauth2.py
```diff
@@ -3,10 +3,33 @@
...
+        'name': 'identity',
+        'description': 'Permission to retrieve your weasyl username and account number',
+    },
+    {
+        'name': 'notifications',
+        'description': 'Access to view your submission inbox, as well as notification counts '
+                       'for comments, favs, messages, etc.',
+    },
+    {
+        'name': 'favorite',
```
strain-113: Inconsistent spelling of favorite with above.

weasyl/errorcode.py
```diff
@@ -100,6 +100,7 @@
         "name correctly and that the account you are trying to recover the password for actually exists."),
     "maxamountInvalid": "The maximum amount you entered is not valid.",
     "minamountInvalid": "The minimum amount you entered is not valid.",
+    "nameMissing": "You did not supply a valid name for this application.",
```
strain-113: nit: `applicationNameMissing` is less ambiguous.

weasyl/oauth2.py
```diff
@@ -3,10 +3,33 @@
 import web
 
 from libweasyl.oauth import get_consumers_for_user, revoke_consumers_for_user, server
+from libweasyl import staff, security
 from weasyl.controllers.base import controller_base
+from weasyl.error import WeasylError
 from weasyl import define as d
 from weasyl import errorcode, login, media, orm
 
+_SCOPES = [
```
strain-113: Why an array of dicts instead of just a dict?

weasyl/oauth2.py
```diff
@@ -82,7 +106,9 @@ def POST(self):
         return body
 
 
-def get_userid_from_authorization(scopes=['wholesite']):
+def get_userid_from_authorization(scopes=None):
+    if not scopes:
+        scopes = ['wholesite']
     valid, request = server.verify_request(*(extract_params() + (scopes,)))
```
strain-113: This is confusing at this point for us. Can we do some more docstrings or comments?

weasyl/oauth2.py
```diff
@@ -92,3 +118,72 @@ def get_userid_from_authorization(scopes=['wholesite']):
...
+    :param name: the name of the application
+    :param scopes: a list of the scopes registered for this application
+    :param redirects: allowed redirect URIs for this application
+    """
+    if not name.strip():
+        raise WeasylError("nameMissing")
+
+    session = d.connect()
+    new_consumer = orm.OAuthConsumer(
+        clientid=security.generate_key(32),
```
strain-113: @charmander probably has better opinions on how to do this properly than I, but since clientid isn't secret, could we make it something more meaningful? maybe a combination of the userid and the timestamp.

weasyl/oauth2.py
```diff
@@ -92,3 +118,72 @@ def get_userid_from_authorization(scopes=['wholesite']):
 __all__ = [
     'get_consumers_for_user', 'revoke_consumers_for_user',
 ]
+
+
+def get_allowed_scopes(user_id):
```
strain-113: style nit: Although it's questionable as a decision, most of the rest of the codebase uses `userid`, and code should be as consistent as possible.

weasyl/oauth2.py
```diff
@@ -92,3 +118,72 @@ def get_userid_from_authorization(scopes=['wholesite']):
...
+         .filter(orm.OAuthConsumer.clientid.in_(clients)))
+    q.delete(synchronize_session=False)
+
+
+def renew_client_secrets(user_id, clients):
+    for cid in clients:
+        q = (orm.OAuthConsumer.query
+             .filter_by(ownerid=user_id)
+             .filter_by(clientid=cid))
+        q.update({orm.OAuthConsumer.client_secret: security.generate_key(64)},
```
strain-113: Just sanity checking: This is one row per run through the loop? And we do it that way so that SQLAlchemy won't use the same security.generate_key(64) result for each?

weasyl/oauth2.py
```diff
@@ -92,3 +118,72 @@ def get_userid_from_authorization(scopes=['wholesite']):
...
+    session = d.connect()
+    new_consumer = orm.OAuthConsumer(
+        clientid=security.generate_key(32),
+        description=name,
+        ownerid=user_id,
+        grant_type="authorization_code",
+        response_type="code",
+        scopes=scopes,
+        redirect_uris=redirects,
+        client_secret=security.generate_key(64),
```
strain-113: nit: Make _SECRET_LENGTH a constant imo.

strain-113:
Ideally this should have unit tests too especially around the oauth2.py methods.

strain-113:
Also a good deal of this, especially the list of scopes and their descriptions and everything new in oauth2.py probably belongs in libweasyl. The existing oauth2.py before this patch consists of little logic and mostly controllers and things to extract information from web requests.

The code should maintain that division where possible: weasyl holding logic specific to the tasks of the 'web' part of 'webapplication' and libweasyl responsible for manipulation of the data models.

weasyl/oauth2.py
taedixon: Not purely insanity, however I believe I may have been overestimating the dangers of mutable default arguments here. It's probably safe to just undo this..

weasyl/oauth2.py
taedixon: That's the intention. I don't know if there's a better / more efficient way of updating keys with unique values through the ORM but there shouldn't be need to reissue more than a few at a time (if any).

templates/oauth2/authorize.html
taedixon: After carefully scouring the [documentation](http://tools.ietf.org/html/rfc6749#section-3.3) it appears that having an application with no scopes is not a valid configuration, so this can just be dropped

templates/control/edit_apikeys.html
taedixon: I agree with the sentiment, however I'm not sure the phrasing is entirely accurate.
> anyone with your client secret effectively has control over your account
Unless I'm mistaken exposing your client secret would actually allow someone to impersonate your application and use it for malicious purposes without being tracable, but a client secret isn't like API keys in giving any special access to the creator's account.

weasyl/oauth2.py
charmander: Maybe you could use a `frozenset`?

templates/control/edit_apikeys.html
charmander: Why? `target="_blank"` is considered irritating to many people and we’d probably want to guard against `window.opener` as a matter of good practice (even though it’d effectively be people attacking themselves here, it’s not too far-fetched to think that we might display integrated apps’ home page somewhere else).

templates/control/edit_apikeys.html
```diff
@@ -97,8 +97,11 @@
         If you are a developer, you may register an application as an OAuth2 consumer.
         You can then use your client id and client secret to request authorization tokens from
         Weasyl. Your client secret MUST be stored securely and should not be accessible
-        by your application's users under any circumstance. For more information see
-        <a href="https://projects.weasyl.com/weasylapi/#oauth2-endpoints">
+        by your application's users under any circumstance. Anyone with your client secret
+        will be able to impersonate you and you will be held responsible for any actions
```
strain-113: Good language.

weasyl/errorcode.py
```diff
@@ -42,6 +42,8 @@
 error_messages = {
     "addressInvalid": "Your IP address does not match the location from which this request was made.",
     "AdminLocked": "This content has been locked from any editing by an admin.",
+    "applicationHasNoScope": "Your OAuth2 application must have at least one access scope.",
```
strain-113: What's your preference for capitalization?

weasyl/oauth2.py
```diff
@@ -26,8 +26,9 @@ def render_form(self, scopes, credentials, mobile, error=None,
         else:
             user = user_media = None
         credentials['scopes'] = scopes
+        detail_scopes = [scope for scope in SCOPES if scope['name'] in scopes]
```
strain-113: Hm. Since it's a sanity error, perhaps a scope not in SCOPES should be an error.

weasyl/errorcode.py
strain-113: Let's start to standardize on UpperCamelCase unless you object.

weykent:
Actually, it seems like #42 should be merged into this one since #40 landed.

charmander:
I had to disable OAuth shortly after #1445 was opened because of a global impersonation bug. It’ll come back eventually, rewritten to be safe, with this functionality included. In the meantime, `X-Weasyl-API` keys will get permission scopes and some other important changes to make them a better stopgap.

PR #18, opened 2016-05-31
Drop support for IE 6–8

charmander:
Makes things cleaner and a bit speedier. Includes an upgrade from jQuery 1.7.2 to 2.2.4.


strain-113:
+1

PR #21, opened 2016-05-31
Remove legacy thumbnail restore tool

charmander:
It’s been around for a year, and that’s probably enough.


strain-113:
Thank you!

PR #22, opened 2016-05-31
Add API methods for journals and characters

Syfaro:
I tried to as closely replicate existing API methods. Let me know if there is anything that needs to be changed.

### DIFF
--- a/weasyl/character.py
+++ b/weasyl/character.py
@@ -24,7 +24,9 @@ import favorite
 from libweasyl import staff, text
 
 from weasyl import macro as m
-from weasyl import api, media
+from weasyl import api
+from weasyl import media
+from weasyl import orm
 
 
 _MEGABYTE = 1048576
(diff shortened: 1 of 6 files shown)


weasyl/character.py
```diff
@@ -24,7 +24,7 @@
 from libweasyl import staff, text
 
 from weasyl import macro as m
-from weasyl import media
+from weasyl import media, api
```
strain-113: Style nit: Lexicographic ordering. Might as well put api on its own line.

weasyl/character.py
```diff
@@ -234,15 +234,63 @@ def select_view(userid, charid, rating, ignore=True, anyway=None):
...
-                                     [charid], ["element"]),
+        "fave_count": favorite.count(charid=charid),
         "comments": comment.select(userid, charid=charid),
         "sub_media": fake_media_items(charid, query[0], login, query[11]),
         "tags": searchtag.select(charid=charid),
     }
 
 
+def select_view_api(userid, charid, anyway=False, increment_views=False):
+    query = define.execute("""
```
charmander: Could you use `define.engine.execute` instead, please? It would look something like this:

```python
query = define.engine.execute("""
    SELECT
        ch.userid, pr.username, ch.unixtime, ch.char_name, ch.age, ch.gender, ch.height, ch.weight, ch.species,
        ch.content, ch.rating, ch.settings, ch.page_views, pr.config
    FROM character ch
        INNER JOIN profile pr USING (userid)
    WHERE ch.charid = %(charid)s
""", charid=charid).fetchone()
```

Then you’ll be able to use `query.settings` instead of `query[11]` and so on below. (We’re trying to get rid of `define.execute`.)

weasyl/journal.py
```diff
@@ -106,14 +106,53 @@ def select_view(userid, rating, journalid, ignore=True, anyway=None):
...
+        'link': d.absolutify_url('/journal/%d/%s' % (journalid, text.slug_for(journal.title))),
+
+        'type': 'journal',
+        'rating': journal.rating.name,
+
+        'views': views,
+        'favorites': favorite.count(journalid=journalid),
+        'comments': comment.count(journalid=journalid),
+        'favorited': favorite.check(userid, journalid=journalid),
+        'friends_only': 'friends-only' in journal.settings
```
charmander: Trailing comma on this line, please.

charmander:
Apart from those two minor bits, this looks good to me!

weasyl/favorite.py
```diff
@@ -254,7 +254,7 @@ def check(userid, submitid=None, charid=None, journalid=None):
         ], options="bool")
 
 
-def count(submitid):
+def count(submitid=None, journalid=None, charid=None):
     return d.execute(
-        "SELECT COUNT(*) FROM favorite WHERE targetid = %i AND type = 's'",
-        [submitid], options=['element'])
+        "SELECT COUNT(*) FROM favorite WHERE targetid = %i AND type = '%s'",
+        [submitid, 's' if submitid else 'f' if charid else 'j'], options=['element'])
```
strain-113: This should be refactored:

1: I'm not sure it does the right thing. It looks like it will always use submitid even if it's None.
2: The chained ifs strike me as a bit of an anti-pattern.
3: It should probably be an error to specify other than exactly one of submitid, journalid, or charid.

Maybe this should be rewritten to take a type and an id. Then ideally the type should not be a single character string (because we want to move away from single character settings) but an enum. Enums, though, are cheap.

Python likes short strings for constant symbols so how about taking an `id` and a `type='submission'` and making it an error if `type` is anything other than `'submission', 'journal'` or `'character'`?

Also should have a doc string for the function.

weasyl/character.py
strain-113: @charmander What's the best way to do this without positional results in query? Is it worth it?

weasyl/journal.py
```diff
@@ -20,7 +20,7 @@
 
 from libweasyl import text, staff
 
-from weasyl import media
+from weasyl import media, orm
```
strain-113: Style nit: I'd prefer separate lines.

weasyl/journal.py
```diff
@@ -106,14 +106,53 @@ def select_view(userid, rating, journalid, ignore=True, anyway=None):
...
-        # todo
-        "fave_count": d.execute("SELECT COUNT(*) FROM favorite WHERE (targetid, type) = (%i, 'j')",
-                                [journalid], ["element"]),
+        "fave_count": favorite.count(journalid=journalid),
         "tags": searchtag.select(journalid=journalid),
         "comments": comment.select(userid, journalid=journalid),
     }
 
 
+def select_view_api(userid, journalid, anyway=False, increment_views=False):
```
strain-113: Too methods here seems redundant. What are the obstacles in combining them?

weasyl/journal.py
```diff
@@ -106,14 +106,53 @@ def select_view(userid, rating, journalid, ignore=True, anyway=None):
...
+    rating = d.get_rating(userid)
+    db = d.connect()
+    journal = db.query(orm.Journal).get(journalid)
+    if journal is None or 'hidden' in journal.settings:
+        raise WeasylError('journalRecordMissing')
+    journal_rating = journal.rating.code
+    if 'friends-only' in journal.settings and not frienduser.check(userid, journal.userid):
+        raise WeasylError('journalRecordMissing')
+    elif journal_rating > rating and userid != journal.userid:
```
strain-113: Consistency nit: The character view api does a lot differently here as well. It checks for staff membership, SFW mode, etc. Is there a reason that they can't have the same behavior?

strain-113:
Good start.
https://pypi.python.org/pypi/sphinxcontrib-napoleon is a good style for docstrings.

weasyl/favorite.py
Syfaro: I just copied this from another function, can definitely improve it!

weasyl/journal.py
Syfaro: Is there a reason for preferring speparate lines? Seems like an unneeded extra line to me.

weasyl/journal.py
Syfaro: This is where a style guide would be really helpful! Easy fix at least :p

weasyl/journal.py
charmander: It makes reordering them easier and insertions or deletions a one-line diff, and stops lines from getting too long (if you add so many to one line that they require wrapping, things start looking less clean).

strain-113: In most IDEs, it also makes an unused warning more visible.

weasyl/favorite.py
strain-113: No worries. The 'obvious' errors are often the ones most easily overlooked. This is why we code review.

strain-113:
Looking at the various `select_view()` versus `select_view_api()` calls, each pair have a lot of repeated logic to retrieve an object, check permissions, etc. Where they differ is in the set of fields returned.

I'm not sure if it's clean, but it might be good to make a try at pulling out the retrieval/checking of a `journal` or `character` orm object into a `_helper_method()` (named something like `_select_journal_and_check()`? Naming is hard) so there isn't the brittleness of two pieces of logic that have to be maintained.

weasyl/journal.py
charmander: Yeah, sorry about that. I started one on the wiki: https://github.com/Weasyl/weasyl/wiki/Style-guide

weasyl/journal.py
Syfaro: Ah, alright! Never thought about those things, thanks for the clarification.

weasyl/favorite.py
```diff
@@ -254,7 +254,27 @@ def check(userid, submitid=None, charid=None, journalid=None):
         ], options="bool")
 
 
-def count(submitid):
+def count(id, contenttype='submission'):
+    """Fetches the count of submissions on some type.
```
charmander: Should this read “Fetches the count of favorites on some type”?

charmander:
+1, thanks!

weasyl/favorite.py
```diff
@@ -254,7 +254,27 @@ def check(userid, submitid=None, charid=None, journalid=None):
...
+    if contenttype == 'submission':
+        querytype = 's'
+    elif contenttype == 'journal':
+        querytype = 'j'
+    elif contenttype == 'character':
+        querytype = 'f'
+    else:
+        raise ValueError("type should be one of 'submission', 'journal', or 'character'")
+
     return d.execute(
```
strain-113: As long as this is changing, can we move this to d.engine.execute() with named parameters?

weasyl/character.py
```diff
@@ -359,10 +406,16 @@ def remove(userid, charid):
     return ownerid
 
 
-def fake_media_items(charid, userid, login, settings):
-    submission_url = define.cdnify_url(
-        define.url_make(charid, "char/submit", query=[userid, settings], file_prefix=login))
-    cover_url = define.cdnify_url(define.url_make(charid, "char/cover", query=[settings], file_prefix=login))
+def fake_media_items(charid, userid, login, settings, absolutify=False):
+    submission_url = define.cdnify_url(define.url_make(charid, "char/submit", file_prefix=login))
```
strain-113: This now ignores its setting parameter, which may break other users. Checking 'Find usages' in pycharm finds select_list() for example.

We may have to spend some time thinking about how to do this right. We could do d.engine.execute() instead of using the nice orm interface and return something more like a raw dict.

We could make define.url_make() take settings in the style of the orm, but that's going to be a lot of code changed in this file too.

We could also do the ugly brittle thing of knowing that `define.url_make` only cares about `file_type` and reconstruct a 'settings' in the old style by reversing the file_type map: https://github.com/Weasyl/libweasyl/blob/master/libweasyl/models/helpers.py#L89
This would work but breaks abstraction barriers.

If this looks like deciding the right thing to do is taking too long, let's separate out the journal endpoint change so it's not blocked on getting this right.

weasyl/comment.py
```diff
@@ -270,6 +270,6 @@ def remove(userid, feature=None, commentid=None):
     return query[1]
 
 
-def count(submitid):
+def count(submitid=None, journalid=None, charid=None):
     # TODO(kailys): refactor to not select from media table
-    return len(select(0, submitid))
+    return len(select(0, submitid=submitid, journalid=journalid, charid=charid))
```
strain-113: There really should be similar ValueError sanity checking here as in favorite's count. Also maybe at the select() level.

By the way, `comment.select()` looks kind of depressing. :\

strain-113:
Nice work on those helper methods. I think journals look in good shape (although I had one small comment about `comment.count()`).

Characters still have some rough edges around the CharSettings columns (it's almost as if a `VARCHAR` of single char flags are a bad way to represent settings....).

As mentioned in another comment, I'd rather we push a good journals endpoint without characters than a good journal endpoint and spaghetti up characters even more in order to give them an endpoint too.

weasyl/character.py
Syfaro: I've fixed the settings param in my local copy which I haven't committed yet. 

Using the orm also breaks a few other things with settings (in views, mostly). I think for the purpose of this PR, it'd be more within scope to revert to the original query and work on that change at a later point.

weasyl/controllers/api.py
```diff
@@ -9,7 +9,7 @@
 from weasyl import define as d, macro as m
 from weasyl import (
     api, character, collection, commishinfo, folder, media, message, profile,
-    index, searchtag, submission, favorite)
+    index, searchtag, submission, journal, favorite)
```
strain-113: Nit: The multiple statements are fine, but they should be ordered (which means fixing index and favorite here).

weasyl/journal.py
```diff
@@ -18,9 +18,12 @@
 import report
 import favorite
 
-from libweasyl import text, staff
+from libweasyl import ratings
```
strain-113: Nit: Ordering.

weasyl/journal.py
```diff
@@ -18,9 +18,12 @@
...
+from libweasyl import text
```
strain-113: Nit: Ordering.

weasyl/journal.py
```diff
@@ -65,52 +68,95 @@ def create(userid, journal, friends_only=False, tags=None):
...
+    Args:
+        userid (int): Currently authenticating user ID.
+        journalid (int): Character ID to fetch.
+        rating (int): Maximum rating to display.
+        ignore (bool): If should respect ignored or blocked tags.
+        anyway (bool): If should ignore checks and display anyway.
+        increment_views (bool): If should increment the number of views on the submission.
+
+    Returns:
+        A journal and all needed data.
```
strain-113: Maybe: A journal and all needed data as a dict.

weasyl/journal.py
```diff
@@ -65,52 +68,95 @@ def create(userid, journal, friends_only=False, tags=None):
...
-def select_view(userid, rating, journalid, ignore=True, anyway=None):
-    journal = d.engine.execute("""
+def _select_journal_and_check(userid, journalid, rating=None, ignore=True, anyway=False, increment_views=True):
+    """Selects a journal, after checking if the user is authorized, etc.
+
+    Args:
+        userid (int): Currently authenticating user ID.
+        journalid (int): Character ID to fetch.
+        rating (int): Maximum rating to display.
+        ignore (bool): If should respect ignored or blocked tags.
```
strain-113: Nit: Include default values. (Updating the style guide for this).

strain-113:
I like how this looks now and I think it made other code cleaner too. +1 but letting charmander take another look and do the merge.

weasyl/character.py
```diff
@@ -186,60 +190,115 @@ def is_hidden(charid):
...
-def select_view(userid, charid, rating, ignore=True, anyway=None):
-    query = define.execute("""
+def _select_character_and_check(userid, charid, rating=None, ignore=True, anyway=False, increment_views=True):
+    """Selects a character, after checking if the user is authorized, etc.
+
+    Args:
+        userid (int): Currently authenticating user ID.
+        charid (int): Character ID to fetch.
+        rating (int): Maximum rating to display.
+        ignore (bool): If should respect ignored or blocked tags.
```
strain-113: Why 'if should' over 'Whether to'? I don't have a problem with it. It's just not something I've seen before.

weasyl/character.py
```diff
@@ -186,60 +190,115 @@ def is_hidden(charid):
...
-        raise WeasylError("characterRecordMissing")
-    elif query[10] > rating and ((userid != query[0] and userid not in staff.MODS) or define.is_sfw_mode()):
-        raise WeasylError("RatingExceeded")
-    elif "f" in query[11] and not frienduser.check(userid, query[0]):
-        raise WeasylError("FriendsOnly")
-    elif ignore and ignoreuser.check(userid, query[0]):
-        raise WeasylError("UserIgnored")
+    elif not query or 'h' in query.settings:
+        raise WeasylError('characterRecordMissing')
+    elif query.rating > rating and ((userid != query.userid and userid not in staff.MODS) or define.is_sfw_mode()):
```
strain-113: I think this might be off wrt define.is_sfw_mode()

Let's say I have SFW mode on, but my default rating is 'Explicit' and an image is mature.

Since mature < explicit, the image will still pass despite SFW mode if I'm reading this if correctly. @charmander is that correct behavior?

weasyl/character.py
Syfaro: No idea, just how I've always written that kind of thing! 

I'll change it when updating the other docstrings.

weasyl/journal.py
Syfaro: What's wrong with the ordering?

weasyl/character.py
charmander: The idea appears to be that mods and the owner should be able to see it regardless of rating unless SFW mode is also on (and that unless `anyway` is `True`). Since it matches the original behaviour, I’d just leave it for now.

strain-113: That's not really the problem, though:

In SFW mode should you ever be able to see mature content?

strain-113: Ah: If SFW mode is on, `rating` will be lower, right? If that's the case, then this makes sense.

weasyl/journal.py
```diff
@@ -65,52 +68,95 @@ def create(userid, journal, friends_only=False, tags=None):
...
+        'content': text.markdown(content),
+        'tags': searchtag.select(journalid=journalid),
+        'link': d.absolutify_url('/journal/%d/%s' % (journalid, text.slug_for(journal.title))),
+        'type': 'journal',
+        'rating': ratings.CODE_TO_NAME[journal.rating],
+        'views': journal.page_views,
+        'favorites': favorite.count(journalid, 'journal'),
+        'comments': comment.count(journalid, 'journal'),
+        'favorited': favorite.check(userid, journalid=journalid),
+        'friends_only': 'f' in journal.settings
```
charmander: Trailing comma!

weasyl/character.py
```diff
@@ -21,10 +21,14 @@
 import report
 import favorite
 
-from libweasyl import staff, text
+from libweasyl import ratings
+from libweasyl import staff
+from libweasyl import text
 
 from weasyl import macro as m
```
charmander: Move this one line down?

strain-113:
+1

charmander:
Looks great; thanks for doing this!

PR #23, opened 2016-06-01
Update library requirements in requirements.txt

skylerbunny:
Updates Weasyl Python library dependencies to current. Depends on libweasyl/#4 and its merge to master being tagged with 0.11.2.


charmander:
Thanks!

PR #28, opened 2016-06-02
Resolve a number of issues introduced with #22

Syfaro:
There were a number of issues with #22 that this should resolve. 

- [x] Missing `posted_at` and `owner_media` field for journal API responses
- [x] Misspelling of table name for comment counts for submissions
- [x] Broken page views for characters and journals
- [x] Missing media ID field (previous API docs say null fields are valid for when URL is unambiguous)

### DIFF
--- a/weasyl/character.py
+++ b/weasyl/character.py
@@ -227,8 +227,10 @@ def _select_character_and_check(userid, charid, rating=None, ignore=True, anyway
     elif ignore and blocktag.check(userid, charid=charid):
         raise WeasylError('TagBlocked')
 
+    query = dict(query.items())
+
     if increment_views and define.common_view_content(userid, charid, 'char'):
-        query.page_views += 1
+        query['page_views'] += 1
 
     return query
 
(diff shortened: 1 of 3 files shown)


weasyl/character.py
```diff
@@ -227,8 +227,10 @@ def _select_character_and_check(userid, charid, rating=None, ignore=True, anyway
     elif ignore and blocktag.check(userid, charid=charid):
         raise WeasylError('TagBlocked')
 
+    query = dict(query.items())
```
charmander: Just `dict(query)` should work.

Syfaro: I had read somewhere that you needed to do `.items()`, suppose I should've tested that for myself. 

Anyway, updated now.

charmander:
Thanks again!

charmander:
(Make sure to run `flake8`, by the way: 3bcf76d)

Syfaro:
Ah, alright. Will do in the future!

PR #29, opened 2016-06-02
Page footer (page_end.html): Update copyright date

kfkitsune:
Changes the year in the copyright line to the current year; previously, the line had 2015 as the end-year.


strain-113:
Looks good. Can you re-issue this with the `Page footer (page_end.html):` removed? Shed painting, but we're trying to maintain imperative voice git summaries.

strain-113:
nbd I guess.

Thanks!

PR #30, opened 2016-06-03
Recenter the position of thumbs

strain-113:
This looks like it was a side-effect of mordernizr being removed and
.csstransform rules no longer being applied.

There's still one more .csstranform rule in the CSS file that I'm
not really sure what to do about.


charmander:
+1

Issue #7, opened 2016-05-30
Update or remove copyright date


taedixon:
Assuming there's a reason for it to appear at the bottom of the page (every website seems to have one), it should probably just get the current year from the system time.

kfkitsune:
This issue can be marked as resolved, as it was fixed in PR #31. The copyright year now automatically updates based on the current year via ``${arrow.now().year}``.

PR #31, opened 2016-06-04
Automatically update year in footer

kfkitsune:
The logical progression from my previous commit which fixed the current year statically; uses datetime to get the year from the system, and passes it to page_end.html. Properly resolves issue #7.

### DIFF
--- a/templates/common/page_end.html
+++ b/templates/common/page_end.html
@@ -1,4 +1,4 @@
-$def with (options=None, active_users_string='')
+$def with (options=None, active_users_string='', current_year='')
   <footer><div id="footer">
 
     <nav role="navigation"><div id="footer-nav" class="clear">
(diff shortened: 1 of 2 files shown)


templates/common/page_end.html
```diff
@@ -65,7 +65,7 @@ <h2 class="typeface-a color-a">Support</h2>
       $if active_users_string:
         ${active_users_string}.
         <br>
-      ${SHA} &middot; Copyright &copy; 2012–2016 Weasyl LLC
+      ${SHA} &middot; Copyright &copy; 2012–${current_year} Weasyl LLC
```
charmander: `${arrow.now().year}` should work and avoid passing more variables to the template.

charmander:
Thanks! Merged as 4004b134590e6807c23ae280dbfce602f3fa7cb6 in 8e83657889082c12c95fee86d401d18cee6963bf.

PR #34, opened 2016-06-05
Add webkit transform prefix for ios and android browsers

taedixon:
reported: https://forums.weasyl.com/vbulletin/showthread.php?8667-KitKat-browser-thumbnail-bug
https://twitter.com/DrKraest/status/739478234371379200

KitKat default browser and iOS <9.2 requires -webkit- prefix to css transform causing problems not remedied by #30


skylerbunny:
+1 : fix tested on Android virtual device.

PR #35, opened 2016-06-05
Hide image and description for Twitter cards on 18+ submissions.

Syfaro:


weasyl/submission.py
```diff
@@ -709,10 +709,23 @@ def twitter_card(submitid):
         content = "[This submission has no description.]"
 
     ret = {
-        'url': d.absolutify_url('/submission/%s/%s' % (submitid, text.slug_for(title))),
-        'description': content,
+        'url': d.absolutify_url(
```
charmander: Is this just a reformatting? (If so, it probably shouldn’t be included.)

weasyl/submission.py
```diff
@@ -709,10 +709,23 @@ def twitter_card(submitid):
...
+            '/submission/%s/%s' % (submitid, text.slug_for(title))),
     }
 
+    if twitter:
+        ret['creator'] = '@%s' % (twitter.lstrip('@'),)
+        ret['title'] = title
+    else:
+        ret['title'] = '%s by %s' % (title, full_name)
+
+    if rating > 20:
```
weykent: The future-proof, non-magic-number way to do this would be using the `Rating.minimum_age` attribute from `libweasyl.ratings`. Something like `if ratings.CODE_MAP[rating].minimum_age >= 18:`

weykent:
Tests? 😺

weasyl/submission.py
Syfaro: Well that's really convenient. But where's the fun in code without magic numbers ;D

weykent:
:+1: looks great. I'd merge it, but my permissions still aren't restored.

PR #37, opened 2016-06-11
Update reCAPTCHA version

kfkitsune:
Code updates to use the newer version of reCAPTCHA. Tests locally with production reCAPTCHA v2 keys work as expected, both with JavaScript and without scripting enabled. (The failure point during tests run on the Weasyl development VM in this situation is the inability to mail anything out. But that is well after any failure mode for the reCAPTCHA code.)

### DIFF
--- a/templates/etc/signup.html
+++ b/templates/etc/signup.html
@@ -14,7 +14,7 @@ $:{RENDER("common/stage_title.html", ["Create a Weasyl Account"])}
     $if captcha_public_key:
       <label for="g-recaptcha-response">Captcha</label>
       <script src="https://www.google.com/recaptcha/api.js" async defer></script>
-      <div class="g-recaptcha" data-sitekey="${captcha_public_key}"></div>
+      <div class="g-recaptcha" data-sitekey="${captcha_public_key}" data-theme="light"></div>
       <noscript>
         <div>
           <div style="width: 302px; height: 422px; position: relative;">
@@ -36,8 +36,6 @@ $:{RENDER("common/stage_title.html", ["Create a Weasyl Account"])}
           </div>
         </div>
       </noscript>
-
-      <p class="color-lighter" style="padding-top: 0.5em;"><i>This helps verify that you are a not a robot.</i></p>
 
     <div class="form-2up clear">
       <div class="form-2up-1">
--- a/templates/etc/signup.html
+++ b/templates/etc/signup.html
@@ -12,19 +12,31 @@ $:{RENDER("common/stage_title.html", ["Create a Weasyl Account"])}
     $ captcha_public_key = CAPTCHA()
 
     $if captcha_public_key:
+      <label for="g-recaptcha-response">Captcha</label>
+      <script src="https://www.google.com/recaptcha/api.js" async defer></script>
+      <div class="g-recaptcha" data-sitekey="${captcha_public_key}"></div>
       <noscript>
-        <iframe src="https://www.google.com/recaptcha/api/noscript?k=${captcha_public_key}"
-                height="300" width="500" frameborder="0"></iframe><br>
+        <div>
+          <div style="width: 302px; height: 422px; position: relative;">
+            <div style="width: 302px; height: 422px; position: absolute;">
+              <iframe src="https://www.google.com/recaptcha/api/fallback?k=${captcha_public_key}"
+                      frameborder="0" scrolling="no"
+                      style="width: 302px; height:422px; border-style: none;">
+              </iframe>
+            </div>
+          </div>
+          <div style="width: 300px; height: 60px; border-style: none;
+                       bottom: 12px; left: 25px; margin: 0px; padding: 0px; right: 25px;
+                       background: #f9f9f9; border: 1px solid #c1c1c1; border-radius: 3px;">
+            <textarea id="g-recaptcha-response" name="g-recaptcha-response"
+                      class="g-recaptcha-response"
+                      style="width: 250px; height: 40px; border: 1px solid #c1c1c1;
+                              margin: 10px 25px; padding: 0px; resize: none;" >
+            </textarea>
+          </div>
+        </div>
       </noscript>
-      <label for="recaptcha_response_field">Captcha</label>
-      <script type="text/javascript"
-              src="https://www.google.com/recaptcha/api/challenge?k=${captcha_public_key}">
-      </script>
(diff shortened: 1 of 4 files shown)


weasyl/controllers/user.py
```diff
@@ -92,19 +92,20 @@ def GET(self):
...
+                "display_reCaptcha": True,
             },
         ])
 
     @define.token_checked
     def POST(self):
         form = web.input(
             username="", password="", passcheck="", email="", emailcheck="",
-            day="", month="", year="", recaptcha_challenge_field="", recaptcha_response_field="")
+            day="", month="", year="", recaptcha_challenge_field="", g_recaptcha_response=web.input("g-recaptcha-response"))
```
weykent: This is a bit.. confused. I can't remember the best way to pull out fields with `-` instead of `_`, but `web.input` in `web.input` is not great.

templates/etc/signup.html
```diff
@@ -12,19 +12,31 @@
     $ captcha_public_key = CAPTCHA()
 
     $if captcha_public_key:
+      <label for="g-recaptcha-response">Captcha</label>
+      <script src="https://www.google.com/recaptcha/api.js" async defer></script>
+      <div class="g-recaptcha" data-sitekey="${captcha_public_key}"></div>
```
weykent: I haven't tried it locally myself yet, but `data-theme="dark"` will probably look better than the light theme. Screenshots of both might be good.

weasyl/controllers/user.py
```diff
@@ -92,19 +92,20 @@ def GET(self):
                 "month": None,
                 "year": None,
                 "error": None,
+                "display_reCaptcha": True,
```
weykent: This doesn't seem to be used?

weykent:
If we're upgrading to v2, using the dedicated [testing key](https://developers.google.com/recaptcha/docs/faq#id-like-to-run-automated-tests-with-recaptcha-v2-what-should-i-do) for local development would be good.

weasyl/define.py
```diff
@@ -329,18 +329,16 @@ def captcha_public():
...
-
-    response = http_post('https://www.google.com/recaptcha/api/verify', data=data)
-    result = response.text.splitlines()
-    return result[0] == 'true'
+        secret=config_obj.get(_captcha_section(), 'private_key'),
+        response=form.g_recaptcha_response,
+        remoteip=get_address())
+    response = http_post('https://www.google.com/recaptcha/api/siteverify', data=data)
+    jsonResult = response.json()
+    return jsonResult['success'][0] == "true"
```
weykent: We've been using `lower_with_underscores` for variable names, but you might as well just write `return response.json()['success'][0] == "true"`

weasyl/define.py
```diff
@@ -337,8 +337,8 @@ def captcha_verify(form):
         response=form.g_recaptcha_response,
         remoteip=get_address())
     response = http_post('https://www.google.com/recaptcha/api/siteverify', data=data)
-    result = response.json()
-    return result['success'] == "true"
+    jsonResult = response.json()
+    return jsonResult['success'][0] == "true"
```
kfkitsune: Slight progress? But not there yet. 

https://gist.github.com/kfkitsune/ea2a1ab08449c7ecc7a4b936e6efaaed

2016-06-11T03:59:37+0000 [stderr#error]   File "/vagrant/weasyl/controllers/user.py", line 105, in POST
2016-06-11T03:59:37+0000 [stderr#error]     if not define.captcha_verify(form):
2016-06-11T03:59:37+0000 [stderr#error]   File "/vagrant/weasyl/define.py", line 341, in captcha_verify
2016-06-11T03:59:37+0000 [stderr#error]     return jsonResult['success'][0] == "true"
2016-06-11T03:59:37+0000 [stderr#error] TypeError: 'bool' object has no attribute '__getitem__'

weasyl/define.py
```diff
@@ -334,12 +334,19 @@ def captcha_verify(form):
...
     response = http_post('https://www.google.com/recaptcha/api/siteverify', data=data)
     jsonResult = response.json()
     print jsonResult
-    return jsonResult['success'] == "true"
+    print jsonResult['success']
+    if jsonResult['success'] == 'True':
+        print 'Yay!'
+        return True
+    else:
+        print "Boo!"
```
kfkitsune: In what negative energy realm does Google return a "True", and when compared to "True" it evaluates as false?

weasyl/define.py
```diff
@@ -329,18 +329,16 @@ def captcha_public():
...
-
-    response = http_post('https://www.google.com/recaptcha/api/verify', data=data)
-    result = response.text.splitlines()
-    return result[0] == 'true'
+        secret=config_obj.get(_captcha_section(), 'private_key'),
+        response=form.g_recaptcha_response['g-recaptcha-response'],
+        remoteip=get_address())
+    response = http_post('https://www.google.com/recaptcha/api/siteverify', data=data)
+    jsonResult = response.json()
+    return jsonResult['success']
```
kfkitsune: I'll clean up this variable name (Line 340/341) tomorrow to match the lower_with_underscore standard and push a commit when done. (Likely also rename it captcha_validation_result, or something slightly more descriptive.)

weykent:
:+1: looks good. Adding tests for this would be icing, and this is going to require a staging deployment to verify it behaves as expected on the live site anyway, so maybe take a look at doing testing in the interim? I'd put it up on staging but my permissions still aren't restored.

weasyl/controllers/user.py
```diff
@@ -99,12 +99,13 @@ def GET(self):
...
             username="", password="", passcheck="", email="", emailcheck="",
-            day="", month="", year="", recaptcha_challenge_field="", recaptcha_response_field="")
+            day="", month="", year="", recaptcha_challenge_field="", 
+            g_recaptcha_response=web.input("g-recaptcha-response"))
 
         if not define.captcha_verify(form):
             return define.errorpage(
                 self.user_id,
-                "The image verification you entered was not correct; you should go back and try again.")
+                "There was an error verifying the data you entered; you should go back and try again.")
```
weykent: Okay, so in trying this out, I realized this error message is way too vague. It should reference the captcha.

weykent:
Awesome! Thanks a bunch. I'm going to squash this because you have a bunch of commits with the same message.

PR #40, opened 2016-06-12
Merge in the libweasyl repo

### LINKED ISSUE #39: Merge libweasyl git repository
libweasyl has been a separate project historically to help with the transition in code styles, etc. as well as to support weasyl3. These considerations are less relevant today however:
 * We're standarizing the codestyle (thanks @charmander and others for the grunt work of fixing 2 space tabs, etc.)
 * weasyl3 isn't exactly abandoned, but our approach is likely to look more incremental going forward.
At the same time juggling two different git projects is complicating code reviews, deployments, etc. Maybe it's time to put the libweasyl code in a libweasyl subdirectory in the weasyl repo. It would also make vagrant provisioning slightly more straightforward and let us use one set of standard logic for tests, coverage, flake8, etc.

n.b. that even if the libweasyl code moves into a sub-directory, it should remain its own python library separate from weasyl itself. Having the data model logic, etc. separate has been helpful for writing things like one off scripts.

weykent:
This also adds travis and codecov. Yaaay!

They can get added to the README later, I think. We should see how well they work in practice first.

Fixes #39.


strain-113:
Is libweasyl/Makefile still necessary? We intend libweasyl and weasyl to share a ve now. If we just had one, top level makefile, it would be one less thing to keep in sync.

weykent:
Good point. I'll nix it.

PR #45, opened 2016-06-13
Don't ignore all of libweasyl

weykent:
Oops! I noticed this as soon as I started trying to change it.


charmander:
Yep, this seems like a good idea.

PR #46, opened 2016-06-13
Use CSS preprocessors

charmander:
This branch doesn’t split the monolithic stylesheet up yet, but it should make doing so easier. A few changes are needed from @Weasyl/ops before this can go on testing and production:

- [x] Add a `gulp sass` step to deployment
- [x] Point `/css` to `build/css` in Nginx configuration with a `Cache-Control` of `public, max-age=31536000`

Saves ~3 KB off the stylesheet size (gzipped) and doesn’t cause it to be redownloaded with every commit unless something actually changed.


weykent:
Wow damn this is great.

weasyl/define.py
```diff
@@ -40,6 +41,9 @@
 
 _shush_pyflakes = [sqlalchemy.orm, config_read]
 
+with open(os.path.join(macro.MACRO_SYS_BASE_PATH, 'build/rev-manifest.json'), 'r') as f:
+    resource_paths = json.loads(f.read())
```
weykent: `json.load(f)` ?

templates/common/page_start.html
```diff
@@ -19,7 +19,7 @@
     <title>Weasyl</title>
 
   <link rel="stylesheet" href="${CDNIFY('/static/fonts/museo500.css')}" />
-  <link rel="stylesheet" href="${CDNIFY('/static/public.css?' + SHA)}" />
+  <link rel="stylesheet" href="${resource_path('css/site.css')}" />
```
strain-113: Should we still append the '?SHA'? The reason we do it is to make the css CDNable.

weasyl/define.py
charmander: I wanted to do this, but anyjson doesn’t seem to provide `load`.

templates/common/page_start.html
charmander: gulp-rev adds the stylesheet’s hash to its filename, which `resource_path` then looks up. (And using the stylesheet’s hash rather than the commit’s means only stylesheet changes will cause it to be redownloaded.)

weasyl/define.py
weykent: Bluuugh I forgot about that. Thanks, anyjson.

templates/common/page_start.html
strain-113: Nice!

assets/scss/site.scss
```diff
@@ -1854,7 +1612,7 @@ blockquote + blockquote {
     height: 1.75rem;
     right: 0;
     top: 0;
-    background: url('./images/icon-delete.png') no-repeat 50% 50%;
+    background: url('../static/images/icon-delete.png') no-repeat 50% 50%;
```
strain-113: Thank you.

PR #50, opened 2016-06-14
Use array-based tag table to improve search performance

charmander:
This might make searches take less than 20 seconds. If it does, we should be able to use it for tag blacklists, too.

### DIFF
--- a/weasyl/search.py
+++ b/weasyl/search.py
@@ -90,7 +90,7 @@ class Query:
             tag = d.get_search_tag(criterion)
             add_nonempty(self.required_includes, tag)
 
-    def __bool__(self):
+    def __nonzero__(self):
         return bool(
             self.possible_includes or
             self.required_includes or
--- a/weasyl/search.py
+++ b/weasyl/search.py
@@ -135,7 +135,7 @@ def select_users(q):
 
 
 def select(userid, rating, limit,
-           search, within, rated, cat, subcat, backid, nextid):
+           search, within, cat, subcat, backid, nextid):
     type_code, type_letter, table, select, subtype = _table_information[search.find]
 
     # Begin statement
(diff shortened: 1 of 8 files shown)


weasyl/search.py
```diff
@@ -99,6 +90,15 @@ def add_nonempty(s, item):
             tag = d.get_search_tag(criterion)
             add_nonempty(self.required_includes, tag)
 
+    def __bool__(self):
```
weykent: Sadly, python 2 and 3 differ here. This needs to be available as both `__bool__` and `__nonzero__`.

weasyl/search.py
```diff
@@ -269,18 +280,33 @@ def make_statement(statement_select, statement_additional_where, statement_order
...
+        "userid": userid,
+        "rating": rating,
+        "ratings": list(search.ratings),
+        "category": cat,
+        "subcategory": subcat,
+        "limit": limit,
+        "backid": backid,
+        "nextid": nextid,
+        "required_include_count": len(search.required_includes),
+    }
```
weykent: Gosh this whole thing makes me sad. Would it be easier to move to query-builder style now? Not that it has to happen in this PR.

charmander: Yes. It’d also be nice to use the same logic for all tables. I’ll tackle both soon =)

weasyl/searchtag.py
```diff
@@ -197,6 +208,17 @@ def associate(userid, tags, submitid=None, charid=None, journalid=None):
                 [table, targetid, d.sql_number_list(list(added))])
 
     if submitid:
+        try:
```
strain-113: Add a TODO here to replace with an upsert when we're on 9.5?

PR #54, opened 2016-06-14
9.5 vagrant

hyena:
This is not ready to be pushed to master. Before it can be pushed we need a plan and timeline around the upgrade path in production.

Putting it up here for comment.


Vagrantfile
```diff
@@ -10,8 +10,11 @@ apt-get install -y ca-certificates apt-transport-https
 
 echo >/etc/apt/sources.list.d/weasyl.list \
     'deb http://apt.weasyldev.com/repos/apt/debian jessie main'
+echo >/etc/apt/sources.list.d/postgresql.list \
+    'deb http://apt.postgresql.org/pub/repos/apt/ trusty-pgdg main 9.5'
```
skylerbunny: This line should be 'deb https://apt.postgresql.org/pub/repos/apt/ jessie-pgdg main' (jessie, not trusty, ant https).

charmander: Any reason we’re not using HTTPS for apt.weasyldev.com, then? (I’m not sure it matters, since everything’s signed anyway.)

skylerbunny:
Weasyl's operations side is now 9.5.3! This can be merged whenever it's ready and reviewed.

skylerbunny:
This installs in Vagrant cleanly from scratch with the 'jessie' to 'trusty' apt repository change. Apt over https is optional and in any case need not block on this being merged.

PR #55, opened 2016-06-14
Improve search performance a bit more

charmander:
- Use an array-based blocktag check (dropped query time from 18s to 3s in one case)
- Stop counting search results after 10,000 (maybe we can stop measuring query time in seconds)

### DIFF
--- a/weasyl/search.py
+++ b/weasyl/search.py
@@ -136,8 +136,8 @@ def select_users(q):
     return ret
 
 
-def select(userid, rating, limit,
-           search, within, cat, subcat, backid, nextid):
+def _find_without_media(userid, rating, limit,
+                        search, within, cat, subcat, backid, nextid):
     type_code, type_letter, table, select, subtype = _table_information[search.find]
 
     # Begin statement
@@ -331,15 +331,6 @@ def select(userid, rating, limit,
         "settings": i.settings,
     } for i in query]
 
-    if search.find == "submit":
-        media.populate_with_submission_media(ret)
-    elif search.find == "char":
-        for r in ret:
-            r["sub_media"] = character.fake_media_items(
-                r["charid"], r["userid"], d.get_sysname(r["username"]), r["settings"])
-    elif search.find == "journal":
-        media.populate_with_user_media(ret)
-
     if backid:
         back_count = d.engine.execute(
             make_statement("SELECT COUNT(*) FROM (SELECT 1", pagination_filter, " LIMIT %(count_limit)s + %(limit)s) _"), params).scalar() - len(ret)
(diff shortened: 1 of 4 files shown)


weasyl/search.py
```diff
@@ -31,6 +31,8 @@
     "journal": (30, "j", "journal", "journalid", 3999),
 }
 
+count_limit = 10000
```
strain-113: Although this is tweaked in tests, it is treated as a constant everywhere else. Should it be COUNT_LIMIT?

weasyl/search.py
```diff
@@ -318,22 +342,22 @@ def get_ids(names):
 
     if backid:
         back_count = d.engine.execute(
-            make_statement("SELECT COUNT(*) FROM (SELECT 1", pagination_filter, ") _"), **params).scalar() - len(ret)
+            make_statement("SELECT COUNT(*) FROM (SELECT 1", pagination_filter, " LIMIT %(count_limit)s + %(limit)s) _"), **params).scalar() - len(ret)
```
strain-113: This seems a counterintuitive use of `%(limit)s`. What's the purpose?

charmander: `len(ret)` is subtracted at the end of the statement, because the count includes results that are part of the current page. Adding `limit` here avoids having “10000+ more” on one side and “9937+ more” on the other.

weasyl/search.py
charmander: Not sure. The capitalization is intentional (I’m less and less sure of the value of formatting constants like that when most variables could be immutable anyway), but I don’t know if it’s the right choice.

weasyl/search.py
```diff
@@ -318,22 +342,22 @@ def get_ids(names):
 
     if backid:
```
strain-113: I thoroughly hate this code. I understand the awkwardness in restructuring it into a if backid: elif nextid: else: tree because of the shared behavior around next_count. But there's too much magic here to understand for someone who didn't write it imo.

For example, why is ret reversed in one case? At over 200 lines of Python SQL logic they'll have long since lost track. At minimum it should have comments.

weasyl/search.py
strain-113: Do you mean it feels wrong to mock when it's UPPERCASE?

weasyl/search.py
```diff
@@ -307,36 +332,71 @@ def get_ids(names):
...
+        # after subtracting if the limit is reached.
         next_count = d.engine.execute(
-            make_statement("SELECT COUNT(*) FROM (SELECT 1", pagination_filter, ") _"), **params).scalar() - len(ret)
+            make_statement("SELECT COUNT(*) FROM (SELECT 1", pagination_filter, " LIMIT %(count_limit)s + %(limit)s) _"),
+            params).scalar() - len(ret)
+
         return ret, next_count, back_count
 
 
+def select(**kwargs):
```
strain-113: Nice.

Issue #47, opened 2016-06-13
Adopt a code of conduct


weykent:
Since this project is open-source now. http://contributor-covenant.org/ is a good template.

strain-113:
I like this a lot. Let's have a chance to review the template before including it though.

weykent:
Absolutely. cc @skylerbunny @ikanimew on setting up an e-mail list for this, and then I'll check in a branch.

skylerbunny:
In general, I like the idea of this project. We ought to discuss it some.

While in the strictest sense this 'waits on ops', the reason an email address is needed is the following, as per the contributor covenant documentation:

_Important! You must add a contact method to the placeholder in the document so that people **know how to report violations.**_ (bold emphasis mine)

Implicit to this is knowing _who_ to report them to, or to put it another way, who is supposed to be on the receiving end of this email address. This is not a technical issue, but a policy question that should be brought up to staff.

This desires discussion both in the 'Do we want to adopt this?' sense, and 'Who in our staff structure deals with violations?' senses.

support@, which already exists, logs a reference-able request to supportdesk and is visible to 'admins+'. It may very well be acceptable for this purpose. Once both points are ironed out, setting up a branch to move forward is pretty straightforward.

weykent:
Cool. Since it's a staff issue, go talk to the staff.

strain-113:
It's tricky because I'd like to distinguish between management of the site and management of the Open Source community. Although that is perhaps an unrealistic separation. Talk about it. In the meantime....

I've mulled over the license and I like it quite a bit. The language strikes a good balance between safe spaces and respecting differing opinions and it maintains a focus on civility. And it's nice that it's becoming an industry standard.

The only caveat I would take is that as our project often deals explicitly with sexual imagery I'd like to clarify that point. Maybe change the first bullet point to just:

 * Unwelcome sexual attention or advances

and add at the bottom of the list language like the following, adopted from http://geekfeminism.wikia.com/wiki/Conference_anti-harassment/Example_exception_to_policy:

Sexual imagery, language, or examples related to site content is permitted as long as (a) it is relevant to the topic of discussion (b) it is presented in a respectful manner, (c) effort has been made to find an alternative.

(c) might be unnecessary there.

Thoughts?

skylerbunny:
codecov@ is now a live mailing list. We can add/remove addresses as we like (and much of this will relate to the discussion above), but it's a valid address!

strain-113:
[...]

PR #57, opened 2016-06-15
Add a CODE_OF_CONDUCT.md

strain-113:
The code of conduct is adapted from http://contributor-covenant.org/.
It has been altered slightly to account for the often explicit
content hosted by this project.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,81 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
+contributors and maintainers pledge to making participation in our project and
+our community a harassment-free experience for everyone, regardless of age, body
+size, disability, ethnicity, gender identity and expression, level of experience,
+nationality, personal appearance, race, religion, or sexual identity and
+orientation.
+
+## Our Standards
+
+Examples of behavior that contributes to creating a positive environment
+include:
+
+* Using welcoming and inclusive language
+* Being respectful of differing viewpoints and experiences
+* Gracefully accepting constructive criticism
+* Focusing on what is best for the community
+* Showing empathy towards other community members
+
+Examples of unacceptable behavior by participants include:
+
+* The use of unwelcome sexual attention or advances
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -68,15 +68,22 @@ the project's `gitter room <https://gitter.im/Weasyl/weasyl>`_.
 
 The above instructions have been tested on Linux and OS X. It should be possible
 to run Weasyl in a virtual environment under Windows, but it hasn't been thoroughly
-tested. 
+tested.
```
charmander: Thank you.

charmander:
Looks good to me.

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,81 @@
...
+Examples of unacceptable behavior by participants include:
+
+* The use of unwelcome sexual attention or advances
+* Trolling, insulting/derogatory comments, and personal or political attacks
+* Public or private harassment
+* Publishing others' private information, such as a physical or electronic
+  address, without explicit permission
+* Other conduct which could reasonably be considered inappropriate in a
+  professional setting
+
```
skylerbunny: I guess this is a nit, but should we put as the last bullet point in the section 'Examples of unacceptable behavior by participants include:'

* The use of sexualized language or imagery, except in discussion of site behavior concerning adult content, as specified immediately below

(or similar, feel free to reword for clarity)

The way it reads now we don't explicitly say 'This is *not* allowed except when'. Adding the above would remove that ambiguity.

With that one possible exception for clarity, I like this and I'm behind it.

weykent:
:+1: modulo the pending issues in #47. I'm not comfortable landing this until those are resolved. cc @iErrored if that gets through!

CODE_OF_CONDUCT.md
strain-113: I'm not sure about that flow. What about if I add an only in here:

Sexual imagery, language, or examples related to site behavior (such as explicit visual submissions that exhibit a bug) are permitted *only* if:

skylerbunny: I think as long as the 'only' (above) is in there, it at least makes clear 'Never allowed except', which is really what I'm looking for. +1.

Issue #43, opened 2016-06-13
configure_libweasyl() shouldn't use a path for staff configuration


weykent:
Currently it takes a path to a file which is a little awkward for deployments like testing.

Reading/parsing a file and its format should be an application level concern and the library should take in a dict or perhaps a named tuple of lists.

strain-113:
This has also made testing more awkward.

charmander:
Fixed by #62,

PR #58, opened 2016-06-16
Permit configure_libweasyl() to use a dict for staff definition

kfkitsune:
Resolves issue #43.

Permits a dict to be passed to ``configure_libweasyl()`` for staff definition, thus eliminating the need for weasyl-staff.yaml. ``weasyl/wsgi.py`` pulls staff definition from ``weasyl/macro.py``, where the default configuration can be defined now (MACRO_SYS_STAFF_CONFIG).

Tests with Travis succeed, and tests in the Weasyl development VM appear to function as expected.


strain-113:
A little bit of historical context that you couldn't have known: this is similar to what it used to look like. In fact, a similar dict was directly in libweasyl originally, so the parameter to `configure_libweasyl` wasn't even there.

The problem with this situation was that whenever the project required a staffing change, we had to cut another release of libweasyl, deploy it, install it, etc. If we had to demote someone from staff quickly, we became blocked on an entire release/deploy process. What we really want is the configuration to be changeable without changing the source code. That way a staff change will not require a new version of Weasyl and a release. Does that make sense?

The easiest way to do this is probably to use the same .yaml file and read its contents in.

kfkitsune:
That makes sense, yes. I have an idea on how that might be done, likely through exposing a Big Red Button to directors (or whichever permission level) that re-calls configure_libweasyl(). I doubt it will be too hard to implement; I'll likely give it a first shot by duplicating the same call that ``weasyl/wsgi.py`` makes. For Issue #43, I'll look into seeing if I can figure out how to get configure_libweasyl to also take a dict for testing; maybe either through making a configure_libweasyl_development(), or an extra parameter that, if set, uses the dict method instead of the YAML file; that is, unless a 'Big Red Button' reload would work as well for testing.

For the sake of a clean(er) commit history, I'll close this PR and remake a new branch for this.

PR #60, opened 2016-06-16
Docker

### LINKED ISSUE #49: Use docker for development/deployment
[Docker](https://www.docker.com/), for all of its pitfalls and hype, might actually fit our use case really well, especially with travis-ci. Travis can build the docker image for the site, and that image can be used for running tests, or development, or production. The web server is nicely self-contained, aside from needing to access the filesystem. Its service dependencies are trivially spun up with `docker compose`.

Developers can also build local images and run those. It's probably even possible to keep the startup time of the development web server as fast as it is already.

This would supersede our current use of Vagrant.

Rollout steps:
- [ ] Docker first pass (#60)
- [ ] Get it running on Windows
- [ ] Use data volumes to keep track of alembic downgraders
- [ ] Build/export docs and assets using Docker.

weykent:
<del>*DO NOT MERGE*. It's not ready.</del>

Fixes #49, whenever this lands.
- [x] Stop using vcversioner or any tag-based versioning system. Apparently it's difficult to ensure that github forks will inherit tags. No idea what this will look like yet.
- [x] <del>Get it running on windows. Windows has been finicky about `docker-compose run`, so something will have to change there.</del> Per @strain-113, I'll do this in a separate PR.
- [x] Test it on linux.
- [x] Write something to drive running `docker-compose` and other commands with the right parameters. Might be easiest to make a little `Dockerfile` in the repo root that's just enough to run the new `wzl` binary. It would have to link in the docker socket.
- [x] Figure out what to do with `.coverage`. It is _so gross_ right now.
- [x] Write docs.


package.json
```diff
@@ -4,6 +4,6 @@
         "gulp-autoprefixer": "^3.1.0",
         "gulp-rename": "^1.2.2",
         "gulp-rev": "^7.0.0",
-        "gulp-sass": "^2.3.1"
+        "gulp-ruby-sass": "^2.0.6"
```
charmander: Why not keep gulp-sass and drop the Ruby dependency?

weykent: I was having trouble installing `gulp-sass` but I can try again. I suppose `gulp-sass` does make it easier to pull down new versions since you don't have to work with both gem and npm.

_wzl/wzl.py
```diff
@@ -0,0 +1,391 @@
...
+))
+@click.argument('args', nargs=-1, type=click.UNPROCESSED)
+@forward_from_wzl_dev
+def test(args):
+    """
+    Run the tests.
+
+    This will also wait until postgres comes up before beginning.
+    """
+
```
charmander: Should this blank line be here? Line 338, too.

weykent: I have mixed feelings about whether to include it or not. I can't decide, so maybe someone else should.

charmander: I don’t see any reason for it to be included. What makes this function different from the others? It’s also against PEP 257.

weykent: The only difference was how I felt when I wrote the functions. I'll remove it.

docker/as-weasyl
```diff
@@ -0,0 +1,16 @@
...
+    olduid="$(id -u weasyl)"
+    if [ "${LOCAL_USER_ID}" != "${olduid}" ]; then
+        echo "=== remapping weasyl (${olduid} => ${LOCAL_USER_ID})"
+        usermod -u "${LOCAL_USER_ID}" weasyl
+        find /weasyl-app /weasyl-src -uid "${olduid}" -print0 \
+            | xargs -r0 chown weasyl
+    fi
+fi
+
+# This MUST be two lines, or sh gets confused.
```
charmander: That’s terrifying

charmander:
The non-documentation parts look good to me. Thanks!

docker/as-weasyl
weykent: Yeah, I have no idea.

weykent:
Okay it's ready.

weykent:
Ping @hyena @skylerbunny about what's necessary to land this. Is it ready?

weykent:
OKAY FINALLY. Travis passed after a ton of spurious failures. Ping again @hyena @skylerbunny.

hyena:
Let's see. Some small busy work and checking with others:
- Remove DO NOT MERGE from the PR text (or cross it out)
- Resolve conflicts. We've been pushing things in under you while this has been out
- Ops needs to weigh in on things. If you and Skyler have been using a separate communications channel for testing this, that's fine. If it's not ready, add the appropriate label
- It's important that others are comfortable with this workflow. So engage with them directly in gitter to try this out
- I haven't been able to run it yet but I don't consider myself a blocker (you caught me at a time when I just turned my dev machine into something that can't do containers) because I'll fix that this weekend

weykent:
- Done.
- You merged things _after_ I posted my comment. :(  I'd been keeping it up to date with conflicts.
- There was a brief conversation but it didn't go anywhere, which is why I've been pinging @skylerbunny. I'll add the label then.
- I think some people did try this out already but I can be pushy about it again.
- Good luck.

hyena:
>You merged things after I posted my comment. :( I'd been keeping it up to date with conflicts.

Those were ready to go.

>I think some people did try this out already but I can be pushy about it again.

I bet they have, but it won't hurt to check.

weykent:
>Those were ready to go.

I know! But I wasn't about to merge every branch into this one just to prevent conflicts.

hyena:
Sorry if that sounded critical. I meant it as more akin to a checklist: "What's the todo list before this can be merged?" The todo list on the PR was the steps needed to get the code complete. It's code complete now, but there's other todos for 'ready to deploy', which is our criteria for merging to master. And I'd like to be intentional with that.

kfkitsune:
Just to track what I identified during the test of the ``wzl`` script the other day, as suggestions:

- Ensure that ``wheelhouse`` and ``node_modules`` directories are not created by root during the ``./wzl build --all`` command. Either pre-create them under the context of the user, or have the script ``chown`` to the user prior to writing to those folders. (Example of such output when it fails: https://paste.weasyldev.com/show/O021ROu7Ig2tP6kMUSIO/ )
- Modify the error message that ``./wzl upgrade-db`` provides when the PSQL service fails to start in the time it expects (and maybe extend the timeout). ``Aborted!`` is not a descriptive error message. (example: https://paste.weasyldev.com/show/m8Ge0VxOIOrMD8HDidTZ/ )
- Maybe add a ``./wzl remove-all`` command in the event a build fails to reset to a known-clean state? (The script knows what images/containers it is making, it should be able to reset the environment as well.) Optionally, confirm if the user intends to execute the command (interactive prompt? ``./wzl remove-all --confirm``?)
- Alter the way ``./wzl setup`` functions to handle ``.stamp-wzl``, or otherwise clean up after itself if a build fails. You (Weykent) suggested having the script check if the image existed as well as if the ``.stamp-wzl`` file exists.
- Maybe have the script check if the user is part of the ``docker`` group and if that is not the case, prompt the user to add themselves to the group (so that running the script as root is not needed since docker would otherwise require the user to be root or ``sudo``). Optionally, provide the commands to the user in the error message for convenience sake.

Other than that, the script worked up to the point I tested it, which was getting the images/containers built and the webserver up and running to the point of being able to log into a user account.

weykent:
- In progress.
- In progress.
- This is a bit dangerous, because if someone is using docker for anything other than weasyl dev, you'd have to be careful about which images get destroyed. I think docker-compose can handle this, though.
- In progress.
- I'm going to try to figure out if there's a way to ask docker "can you connect to the docker socket?" because the docker group is an implementation detail of linux. OS X doesn't have a docker group at all, because it uses docker-machine. Troubleshooting "can't connect to the docker socket" errors is going to be OS-specific.

kfkitsune:
> This is a bit dangerous, because if someone is using docker for anything other than weasyl dev, you'd have to be careful about which images get destroyed. I think docker-compose can handle this, though.

Perhaps add an environment variable check, or other mechanism for ``./wzl`` to detect that the environment isn't development? Such as via a bash ``export WEASYL_PRODUCTION=true`` or something. Even say, a directory at ``~/.wzl-is-production``. Basically, marking some method of safeguarding the more critical containers (database, media, etc.).

I saw from the QuantifiedCode rules that environment variables can be read via Python, so that is why I suggest that bit. It's not critical if you don't want to directly expose a command like that. Perhaps note the commands in the documentation?

weykent:
Sorry, to be clear, I meant "weasyl dev" as opposed to "`{any other project}` dev", not as opposed to "weasyl production". We are not literally the only project in existence using docker. 😛

hyena:
It might be possible to do a 'make dist-clean' equivalent in wzl that operates in terms of filesystem rather than docker objects. In a vagrant parallel, do a `rm -rf .vagrant` instead of `vagrant destroy`. Normally that sort of thing is gross. However, it might be worth considering if we're worried about dangerous docker commands.

hyena:
Had a chance to try this out finally. The `wzl` script is pretty sleek. Some issues:
 - [ ] The instructions should mention under requirements that the `wzl` script requires that the user be a member of the `docker` group to run correctly.
 - [ ] The requirements should mention that the images will consume about ____ gigabytes of space. It looked to be about 3 to me from a glance at `docker images`
 - [ ] Several files are currently created as root on a fresh install by these instructions: `libweasyl/libweasyl/_version.py`, `wheelhouse`, the `site-config.txt`, etc.
 - [ ] `wzl url` doesn't appear to work with `wzl run -n`
 - [ ] `wzl` doesn't appear to have a command to stop the containers.

weykent:
_IT'S BACK_

Okay, going through the checklists now.

weykent:
>Ensure that `wheelhouse` and `node_modules` directories are not created by root during the `./wzl build --all` command. Either pre-create them under the context of the user, or have the script `chown` to the user prior to writing to those folders. (Example of such output when it fails: https://paste.weasyldev.com/show/O021ROu7Ig2tP6kMUSIO/ )

I didn't add an explicit check for this but I can't reproduce it anymore. I might've fixed it? Maybe ephemeral? Everyone should try this out again anyway.

>Modify the error message that `./wzl upgrade-db` provides when the PSQL service fails to start in the time it expects (and maybe extend the timeout). `Aborted!` is not a descriptive error message. (example: https://paste.weasyldev.com/show/m8Ge0VxOIOrMD8HDidTZ/ )

I had _tried_ to make it descriptive, but I fixed it so the message is actually shown now.

>Maybe add a `./wzl remove-all` command in the event a build fails to reset to a known-clean state? (The script knows what images/containers it is making, it should be able to reset the environment as well.) Optionally, confirm if the user intends to execute the command (interactive prompt? `./wzl remove-all --confirm`?)

>It might be possible to do a 'make dist-clean' equivalent in wzl that operates in terms of filesystem rather than docker objects. In a vagrant parallel, do a `rm -rf .vagrant` instead of `vagrant destroy`. Normally that sort of thing is gross. However, it might be worth considering if we're worried about dangerous docker commands.

`./wzl clean` and `./wzl clean --dist` should handle all this now.

>Alter the way `./wzl setup` functions to handle `.stamp-wzl`, or otherwise clean up after itself if a build fails. You (Weykent) suggested having the script check if the image existed as well as if the `.stamp-wzl` file exists.

I found a better alternative to crappy stamp files: ask docker directly.

Last bits remaining, as I'm going to bed now: docs. I'll tick off the checkboxes when those are done too.

weykent:
Oh, and I guess travis broke too.

weykent:
Okay, docs updated. Going to fix the test failures upstream and merge them in.

weykent:
Hooray, tests properly pass again.

hyena:
Nice work on the cleaning commands.

1. On Ubuntu 16.04 with docker installed via `apt-get install docker.io` and `hyena` added to the docker group, I'm still getting some odd permissions:
  * wheelhouse is owned by `hyena root`. Its contents are owned by `hyena 65533`. I don't believe 65533 is valid on my system.
  * node_modules are still owned by `root root`, though its contents is empty.
2. `./wzl build --all` dies while doing the sass work: https://paste.weasyldev.com/show/K978PgpfkzzYpgtvhb1L/

The two may be related if files aren't available due to permissions.

weykent:
@hyena did you do a `./wzl clean --dist` before the `build --all`? Anything you had before need to be cleared out.

hyena:
This was on a completely fresh checkout.

weykent:
@hyena sadly, 'fresh checkout' implies nothing about the state of docker.

weykent:
Permissions/ownership issues should all be fixed.

weykent:
Okay, and this should solve the last lingering `node_modules` issues.

docker/alpine-3.5-repositories
```diff
@@ -0,0 +1,2 @@
+http://dl-cdn.alpinelinux.org/alpine/v3.5/main
```
charmander: What does this file do?

weykent: Good catch! Nothing, anymore. It used to be for forcing npm's alpine image from 3.4 to 3.5, but nothing uses npm's image now.

skylerbunny:
For reasons I haven't quite nailed down yet, running './wzl build --all' fails when attempted. (This happens either with the branch as it stands, or if it's been merged to current master with conflicts resolved.)

It's probably as simple as a build order dependency issue, becuase I can work around this problem by first issuing `./wzl build weasyl-app` and THEN `./wzl build --all'`.

```
==> ['docker-compose', '-f', 'docker-compose.yml', 'up', '-d', 'db']
Creating weasylsrc_db_1 ...
Creating weasylsrc_db_1 ... done
WARNING: Image for service db was built because it did not already exist. To rebuild this image you must use `docker-compose build` or `docker-compose up --build`.
==> ['docker-compose', 'run', 'weasyl-app-dev', 'alembic', '--', 'upgrade', 'head']
Starting weasylsrc_db_1 ... done
Creating weasylsrc_cache_1 ...
Creating weasylsrc_cache_1 ... done
Building weasyl-app-dev
Step 1/8 : FROM weasyl-app
ERROR: Service 'weasyl-app-dev' failed to build: pull access denied for weasyl-app, repository does not exist or may require 'docker login'
```

PR #65, opened 2016-06-17
Make 'make-test' work when pytest or flake8 not installed

skylerbunny:
Installs both the python libraries pytest and flake8 when 'make test' is run and neither of these libraries are installed in weasyl-env.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -108,6 +108,7 @@ guest-run: .vagrant
 # Phony target to run tests
 .PHONY: test
 test: setup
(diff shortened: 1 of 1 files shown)


charmander:
How about in the `$(VE)` target?

skylerbunny:
I've moved this to the $(VE) target as suggested. It looks like a better place for it to live, and avoids having it be run on every 'make test', also a good thing.

charmander:
Thanks!

PR #67, opened 2016-06-17
Remove SQL formatting functions that are or should be unused

charmander:
A little more progress towards parameters everywhere. Includes bonus tag tests.


weasyl/define.py
```diff
@@ -166,38 +166,7 @@ def sql_number_list(target):
...
-
-    return "(%s)" % (", ".join(["'%s'" % (sql_escape(i)) for i in target]))
-
-
-def sql_string_series(target):
-    if not target:
-        raise ValueError
-
-    return ", ".join(sql_string_list(i) for i in target)
+    return "(%s)" % (", ".join(["%d" % (i,) for i in target]))
```
weykent: How many places is this still used?

charmander: 11\. I suppose you want me to get rid of those =)

weykent: Nah. Incremental is good. I was just hoping we'd coincidentally obsoleted this function.

Issue #70, opened 2016-06-17
Add staff action audit log


ikanimew:
We already track staff actions by using staff notes. Create a page within the admin tools that lists all staff actions, with filtering capabilities.

Ideally filter options would be text fields for username and staff member, as well as date ranges (perhaps default to 30 days, so the initial page doesn't have a huge load), and action type.

Staff notes are stored in the comments table, with settings ~ 's'.

strain-113:
Although multi-faceted, this should be relatively straightforward. I'm marking this new dev friendly.

weykent:
I suppose it's out of scope to suggest that in the newfound spirit of transparency, this should be made publicly-viewable like [Something Awful's leper colony](http://forums.somethingawful.com/banlist.php).

ikanimew:
Our code is transparent. Our user information is not, and should not be. This was discussed way back at the beginning of the site IIRC and we came to the conclusion that a demonstration of what's recorded would be fine ("This is what we to to track and audit staff actions internally"), but not a display of actual data.

PR #73, opened 2016-06-17
Add a staff action audit log

Syfaro:
For #70, adds a staff action audit log. 

screenshot 2016-06-17 18 34 39


weasyl/moderation.py
```diff
@@ -717,3 +717,54 @@ def note_about(userid, target_user, title, message=None):
             userid=userid, target_user=target_user, unixtime=arrow.utcnow(),
             settings='s', content=staff_note,
         ))
+
+
+def audit_log(username=None, staff=None, start_date=None, end_date=None):
+    sql = """
```
charmander: The SQLAlchemy query builder should probably be used for this.

templates/modcontrol/auditlog.html
```diff
@@ -0,0 +1,46 @@
+$def with ( log, form={} )
```
charmander: If I recall correctly, other templates don’t have spaces inside the parentheses here.

templates/modcontrol/auditlog.html
```diff
@@ -0,0 +1,46 @@
...
+      <label for="staff">Staff member</label>
+      <input type="text" name="staff" class="input" id="staff" value="$form['staff']">
+
+      <label for="start_date">Start date</label>
+      <input type="datetime" name="start_date" class="input" id="start_date" value="$form['start_date']">
+
+      <label for="end_date">End date</label>
+      <input type="datetime" name="end_date" class="input" id="end_date" value="$form['end_date']">
+
+      <button type="submit" class="button positive">Filter</button>
```
charmander: `type="submit"` is the default.

templates/modcontrol/auditlog.html
```diff
@@ -0,0 +1,46 @@
...
+            <a href="/~$event['user_name']">$event['mod_name']</a> &middot; <a href="/modcontrol/auditlog?staff=$event['user_name']">All</a>
+          </td>
+          <td>$:event['content']</td>
+          <td>${DATE(event['unixtime'])} at ${TIME(event['unixtime'])}</td>
+        </tr>
+    </tbody>
+  </table>
+
+  <div class="form skinny clear">
+    <form method="GET" action="/modcontrol/auditlog" class="clear">
```
charmander: Do both of these elements need `clear`? Does either of them?

templates/modcontrol/modcontrol.html
```diff
@@ -2,6 +2,9 @@
 
 <div id="modcontrol_content" class="content" style="padding-top:2em;">
   <div class="form skinny">
+    <p style="font-size:1.3em"><a href="/modcontrol/auditlog">Audit Log</a></p>
```
charmander: Inline styles are bad. Maybe we can fix that for the rest of this page? (Plus, these aren’t paragraphs. Well, if not, at least use a semicolon for consistency.)

weasyl/controllers/moderation.py
```diff
@@ -245,3 +245,17 @@ def POST(self):
...
+class modcontrol_auditlog_(controller_base):
+    login_required = True
+    moderator_only = True
+
+    def GET(self):
+        form = web.input(
+            username=None, staff=None, start_date=None, end_date=None)
+
+        return define.webpage(self.user_id, "modcontrol/auditlog.html", [
+            moderation.audit_log(**form),
```
charmander: Note that `form` includes everything POSTed; the arguments to `web.input` are just defaults. It might be best to pass each argument individually here.

templates/modcontrol/auditlog.html
Syfaro: Yep, you are correct. Not sure how that ended up there.

templates/modcontrol/auditlog.html
Syfaro: Nope, will remove.

templates/modcontrol/modcontrol.html
Syfaro: Removed all inline styles on the page and switched to those pages being a list.

Hendikins:
Had a look at this on testing, and it's displaying oldest items first. For consistency with the rest of the mod tools it would be better if newest items were displayed first.

weasyl/moderation.py
```diff
@@ -719,52 +719,88 @@ def note_about(userid, target_user, title, message=None):
...
+        An int.
+    """
+
+    db = d.connect()
+
+    user = d.meta.tables['login']
+
+    query = d.sa.select([
+        user.c.userid,
+    ]).where(user.c.login_name.op('~')(login_name))
```
charmander: This doesn’t seem to match the documentation.

weasyl/moderation.py
```diff
@@ -719,52 +719,88 @@ def note_about(userid, target_user, title, message=None):
...
+    db = d.connect()
+
+    comment = d.meta.tables['comments']
+    user = d.meta.tables['login']
+
+    query = d.sa.select([
+        comment.c.commentid,
+        comment.c.content,
+        comment.c.unixtime,
+        (d.sa.select([user.c.login_name]) \
```
charmander: No backslashes necessary.

weasyl/moderation.py
```diff
@@ -719,52 +719,88 @@ def note_about(userid, target_user, title, message=None):
...
+             .select_from(user) \
+             .where(comment.c.target_user == user.c.userid)).label('user_name'),
+        (d.sa.select([user.c.login_name]) \
+             .select_from(user) \
+             .where(comment.c.userid == user.c.userid)).label('mod_name'),
+    ]).where(comment.c.settings.op('~')('s')) \
```
charmander: Except for here, where you should wrap the whole thing in parentheses instead to avoid the backslash.

weasyl/moderation.py
Syfaro: What doesn't seem to match? Given a login name, it selects the user's ID.

charmander: It appears to take a regular expression to match a login name, not a login name.

weasyl/moderation.py
charmander: Also, it’s still possible to use joins with the query builder, which would save introducing `get_userid_from_loginname` and the extra queries and subqueries.

weasyl/moderation.py
Syfaro: Is there some reason this is preferred? More parentheses seems more complicated to understand.

weasyl/moderation.py
Syfaro: I was having a really hard time figuring out how to use joins in it and found that elsewhere, would it be possible to get an example of properly using joins here?

weasyl/moderation.py
Syfaro: Ah, alright. Not really sure why it's using a regex in the first place.

weasyl/moderation.py
charmander: What it looks like with backslash-avoiding parentheses in place, redundant parentheses removed, the imported `sa`, and indentation instead of alignment:

```python
query = (
    sa.select([
        comment.c.commentid,
        comment.c.content,
        comment.c.unixtime,
        sa.select([user.c.login_name])
            .select_from(user)
            .where(comment.c.target_user == user.c.userid)
            .label('user_name'),
        sa.select([user.c.login_name])
            .select_from(user)
            .where(comment.c.userid == user.c.userid)
            .label('mod_name'),
    ])
        .where(comment.c.settings.op('~')('s'))
        .order_by(comment.c.commentid.desc())
)
```

weasyl/moderation.py
charmander:
```python
comments = d.meta.tables['comments']
moderator = d.meta.tables['login'].alias('m')
user = d.meta.tables['login'].alias('u')

query = (
    sa
        .select([
            comments.c.commentid,
            comments.c.content,
            comments.c.unixtime,
            moderator.c.login_name.label('mod_name'),
            user.c.login_name.label('user_name')])
        .select_from(
            comments
                .join(moderator, comments.c.userid == moderator.c.userid)
                .join(user, comments.c.target_user == user.c.userid))
        .where(comments.c.settings.op('~')('s'))
        .order_by(comments.c.commentid.desc())
)
```

Syfaro: Ah, thank you!

charmander:
Okay! The code looks good to me now; seems we’re just looking for moderators’ opinions on the UI.

weasyl/controllers/moderation.py
```diff
@@ -17,6 +17,19 @@ def modcontrol_(request):
 
 
 @moderator_only
+def modcontrol_auditlog_(request):
+    form = request.web_input(username=None, staff=None, start_date=None, end_date=None)
+
```
charmander: There’s whitespace here; flake8 should report this sort of thing. (We ought to run it in addition to tests and coverage on pull requests, really.)

weasyl/moderation.py
```diff
@@ -718,3 +718,80 @@ def note_about(userid, target_user, title, message=None):
...
+
+    db = d.connect()
+
+    comments = d.meta.tables['comments']
+    moderator = d.meta.tables['login'].alias('m')
+    user = d.meta.tables['login'].alias('u')
+
+    query = (
+        sa
+            .select([
```
charmander: That being said, flake8 doesn’t like this, but it’s wrong. =)

templates/modcontrol/auditlog.html
```diff
@@ -0,0 +1,48 @@
...
+          <td>$:event['content']</td>
+          <td>${DATE(event['unixtime'])} at ${TIME(event['unixtime'])}</td>
+        </tr>
+    </tbody>
+  </table>
+
+  <div class="form skinny">
+    <form method="GET" action="/modcontrol/auditlog">
+      <label for="username">Username</label>
+      <input type="text" name="username" class="input" id="username" value="$form['username']">
```
charmander: There’s already an element with the id “username” in the header. For now, just picking a non-colliding id is probably fine; I’ll change it to use `<span class="label">` later on.

templates/modcontrol/auditlog.html
```diff
@@ -0,0 +1,48 @@
+$def with (log, form={})
```
charmander: Does `form` need a default here?

weasyl/moderation.py
```diff
@@ -718,3 +718,80 @@ def note_about(userid, target_user, title, message=None):
...
+        username (str): Login name to filter by actions done to. Defaults to None.
+        staff (str): Login name to filter by actions done by. Defaults to None.
+        start_date (str): Show events after this date.
+            Accepts any string that arrow.get can parse. Defaults to None.
+        end_date (str): Show events before this date.
+            Accepts any string that arrow.get can parse. Defaults to None.
+        timezone (str): Timezone used to parse provided dates.
+
+    Returns:
+        A list of dicts with event data. Content is Markdown formatted.
```
charmander: This is easy to misinterpret. Maybe “Content is rendered to HTML from Markdown” or something?

(Also, while you’re at it, please remove the blank lines after both docstrings.)

weasyl/moderation.py
```diff
@@ -718,3 +718,80 @@ def note_about(userid, target_user, title, message=None):
...
+            start_date = start_date.replace(tzinfo=timezone)
+
+        query = query.where(comments.c.unixtime > start_date)
+
+    if not end_date or not start_date:
+        query = query.limit(100)
+
+    events = []
+
+    for event in db.execute(query).fetchall():
```
charmander: Does `d.engine` work in place of `db`?

assets/scss/site.scss
```diff
@@ -3926,6 +3926,59 @@ img + #detail-art-text {
...
+        float: right;
+        margin-top: 1em;
+    }
+
+    .help-text {
+        padding: 10px 0;
+    }
+}
+
+.modcontrol {
```
charmander: Are these `.modcontrol` styles used?

PR #74, opened 2016-06-18
Speed up Recently Popular calculation

charmander:
Taking advantage of the fact that you can’t favorite something before it’s created, this reduces the number of favorite rows read from 5.5 million to a few thousand. (It’s possible to do the same thing without adding an index, but this makes the query neater.)

### DIFF
--- a/weasyl/blocktag.py
+++ b/weasyl/blocktag.py
@@ -54,7 +54,7 @@ def check(userid, submitid=None, charid=None, journalid=None):
 
 
 def check_list(rating, tags, blocked_tags):
-    return any(rating >= b['rating'] and b['title'] in tags for b in blocked_tags)
+    return any(rating >= b['rating'] and b['tagid'] in tags for b in blocked_tags)
 
 
 def suggest(userid, target):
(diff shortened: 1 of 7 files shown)


weasyl/blocktag.py
```diff
@@ -54,7 +54,7 @@ def check(userid, submitid=None, charid=None, journalid=None):
 
 
 def check_list(rating, tags, blocked_tags):
-    return any(rating >= b['rating'] and b['title'] in tags for b in blocked_tags)
```
strain-113: :\

weasyl/submission.py
```diff
@@ -1053,41 +1051,32 @@ def select_recently_popular():
...
-        'subtype': s.subtype,
-        'unixtime': s.unixtime.timestamp,
-        'tags': list(s.tags),
-        'userid': s.userid,
-        'username': p.username,
-    } for s, p, sc in q.all()]
+
+    query = d.engine.execute("""
+        SELECT
+            log(count(favorite.*) + 1) + log(submission.page_views + 1) / 2 + submission.unixtime / 180000 AS score,
```
strain-113: I know the accuracy loss is minor, but how much savings does this have over the max approach?

weasyl/blocktag.py
charmander: The front page was the only thing using `check_list` and `cached_select`, so…!

weasyl/submission.py
strain-113: Nit: Update the docstring to reflect the new calculation.
Nit: formatting or use of ()'s to make the order of operations unambiguous.

Issue #78, opened 2016-06-20
Adapt to Raven 5.x API change


skylerbunny:
The following errors get logged when our code uses a 5.x (relatively new) version of Raven.  We are currently on 5.0.21. In production these get logged every three minutes.

```
2016-06-20_19:55:01.56840 2016-06-20T12:55:01-0700 [-] refreshed active user counts
2016-06-20_19:58:51.25545 2016-06-20T12:58:51-0700 [twisted.web.client._HTTP11ClientFactory#info] Starting factory <twisted.web.client._HTTP11ClientFactory instance at 0x7fb3389e1dd0>
2016-06-20_19:58:51.25615 2016-06-20T12:58:51-0700 [wsgi] Traceback (most recent call last):
2016-06-20_19:58:51.25617 	  File "/var/lib/weasyl/weasyl/weasyl-env/local/lib/python2.7/site-packages/web/application.py", line 239, in process
2016-06-20_19:58:51.25617 	    return self.handle()
2016-06-20_19:58:51.25617 	  File "/var/lib/weasyl/weasyl/weasyl/middleware.py", line 106, in weasyl_exception_processor
2016-06-20_19:58:51.25618 	    event_id, = web.ctx.env['raven.captureException'](request_id=request_id)
2016-06-20_19:58:51.25618 	ValueError: too many values to unpack
2016-06-20_19:58:51.25618
2016-06-20_19:58:51.25640 2016-06-20T12:58:51-0700 [wsgi]
2016-06-20_19:58:51.25641
```

It is currently talking to Sentry 8.5.1.

weykent:
https://github.com/getsentry/raven-python/commit/8c89e45a68135095f130ad8a9c1a71cfe5c67b5f Looks like it now returns just the `event_id`, without being wrapped in a one-tuple.

skylerbunny:
From https://github.com/getsentry/raven-python/blob/master/CHANGES :

Version 5.2.0: Protocol version is now 6 (requires Sentry 7.0 or newer).

PR #79, opened 2016-06-21
Sentry error-reporting improvements

charmander:
- Restores friendly error pages by fixing #78
- Strips session ids from events
- Uses built-in fields for event data when available so Sentry can add useful widgets

Tested using a local Sentry instance.

### DIFF
--- a/weasyl/middleware.py
+++ b/weasyl/middleware.py
@@ -166,7 +166,7 @@ class SentryEnvironmentMiddleware(object):
         self.app = app
         self.client = raven.Client(
             dsn=dsn,
-            release=raven.fetch_git_sha(os.path.abspath(os.path.join(__file__, '../..'))),
+            release=d.CURRENT_SHA,
             processors=[
                 'raven.processors.SanitizePasswordsProcessor',
                 'weasyl.middleware.RemoveSessionCookieProcessor',
--- a/weasyl/middleware.py
+++ b/weasyl/middleware.py
@@ -103,7 +103,7 @@ def weasyl_exception_processor():
         request_id = None
         if 'raven.captureException' in web.ctx.env:
             request_id = base64.b64encode(os.urandom(6), '+-')
-            event_id, = web.ctx.env['raven.captureException'](request_id=request_id)
+            event_id = web.ctx.env['raven.captureException'](request_id=request_id)
             request_id = '%s-%s' % (event_id, request_id)
         print 'unhandled error (request id %s) in %r' % (request_id, web.ctx.env)
         traceback.print_exc()
--- a/weasyl/middleware.py
+++ b/weasyl/middleware.py
@@ -19,14 +19,9 @@ from libweasyl import security
 from libweasyl.cache import ThreadCacheProxy
 from weasyl import define as d
 from weasyl import errorcode
+from weasyl import http
 from weasyl import orm
 from weasyl.error import WeasylError
-
-
-def _get_headers(env):
-    return {
-        key[5:].replace('_', '-').title(): value
-        for key, value in env.iteritems() if key.startswith('HTTP_')}
 
 
 class ClientGoneAway(Exception):
@@ -188,7 +183,7 @@ class SentryEnvironmentMiddleware(object):
                 'method': web.ctx.env['REQUEST_METHOD'],
                 'data': web.webapi.rawinput(method='POST'),
                 'query_string': web.ctx.env['QUERY_STRING'],
-                'headers': _get_headers(web.ctx.env),
+                'headers': http.get_headers(web.ctx.env),
                 'env': web.ctx.env,
             },
         }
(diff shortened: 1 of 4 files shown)


weasyl/middleware.py
```diff
@@ -19,6 +23,12 @@
 from weasyl.error import WeasylError
 
 
+def _get_headers(env):
+    return {
+        key[5:].replace('_', '-').title(): value
+        for key, value in env.iteritems() if key.startswith('HTTP_')}
```
weykent: God it makes me sad that `web.py` doesn't have this. I was _pretty_ sure it existed somewhere else in the project already, because `oauthlib` needed the same thing.

weasyl/middleware.py
```diff
@@ -132,30 +163,52 @@ def __call__(self, environ, start_response):
 
 class SentryEnvironmentMiddleware(object):
     def __init__(self, app, dsn, reactor=None):
-        import raven
         self.app = app
-        self.client = raven.Client(dsn)
+        self.client = raven.Client(
+            dsn=dsn,
+            release=raven.fetch_git_sha(os.path.abspath(os.path.join(__file__, '../..'))),
```
weykent: When would this be different from `define.CURRENT_SHA`? Or does `fetch_git_sha` do zany magic and/or something else?

weasyl/middleware.py
```diff
@@ -132,30 +163,52 @@ def __call__(self, environ, start_response):
...
+            processors=[
+                'raven.processors.SanitizePasswordsProcessor',
```
weykent: So this was enough to catch the password confirmation fields too?

weasyl/middleware.py
```diff
@@ -132,30 +163,52 @@ def __call__(self, environ, start_response):
...
+        data = {
+            'level': level,
+            'user': {
+                'id': d.get_userid(),
+                'ip_address': d.get_address(),
+            },
+            'request': {
+                'url': web.ctx.env['PATH_INFO'],
+                'method': web.ctx.env['REQUEST_METHOD'],
+                'data': web.webapi.rawinput(method='POST'),
```
weykent: This makes me suspicious. Does `web.py` cache all of the POST request body? i.e. is this safe to call after `web.input` has been called a few times? Expect the unexpected and all.

weasyl/middleware.py
weykent: By the way, it's not clear to me whether the processors here are intended to make the things on sentry safe enough that sentry can be public-ish or if sentry should be locked down. These processors are definitely not enough to prevent sentry from capturing PII and other sensitive information that can't necessarily _directly_ be used to compromise an account.

weasyl/middleware.py
charmander: @weykent: Ah, I guess this does depend on `define`. The only difference is that `fetch_git_sha` is the full hash; `CURRENT_SHA` is a shorthash.

weasyl/middleware.py
weykent: Yep, [it's in `weasyl.oauth2`](https://github.com/Weasyl/weasyl/blob/4ad53b59da2fd7452546f04025414780911190f2/weasyl/oauth2.py) as `extract_params`.

weasyl/middleware.py
charmander: Yes.

weasyl/middleware.py
charmander: Yes, it does – even for multipart/form-data. (Uploads make the requests non-replayable, but that’s really the best solution.)

weasyl/middleware.py
weykent: Is a shorthash fine? I don't have a local sentry to test again. I don't even know where this gets used in its UI.

charmander: `release` is an arbitrary identifier, so a shorthash is definitely distinct for each deployment and probably fine.

weasyl/middleware.py
charmander: Where would be a good place to import it from in each?

weykent: Would you hate me too much for suggesting putting it in the `web.py` fork?

Trying to prevent enabling people who think `web.py` is a good idea aside, maybe `weasyl.compat` or `weasyl.dry`. There's a few dumping grounds for things like this. (Maybe worth a cleanup ticket? Not to sidetrack this too much.)

weasyl/middleware.py
charmander: > So this was enough to catch the password confirmation fields too?

Yes.

> By the way, it's not clear to me whether the processors here are intended to make the things on sentry safe enough that sentry can be public-ish or if sentry should be locked down.

This is definitely not with the intent of making Sentry public. Session ids probably even still show up in locals somewhere.

weykent: :+1: on that, then.

Issue #51, opened 2016-06-14
Switch to postgres 9.5+ upsert


weykent:
[Postgres 9.5 and greater have upsert](https://www.postgresql.org/docs/9.5/static/sql-insert.html), which would be great to use instead of catching constraint violations. This also makes it easier to move to using transactions in more places.

Waiting on a 9.5 deploy.

strain-113:
More than just the deploy, I'd like to see a plan around the database upgrade and what downtime if any it might require.

skylerbunny:
Well, I feel like a heel, seeing as this ticket asked for the deploy plan as well as the deploy itself (I missed that part of this request).

I did in any case try to keep our gitter channel up to date about the upgrade to db1 in production today. I realize this will be a post-mortem now rather than the advance plan, but it is what I planned in advance and the list I worked off of, expanded a bit for readability.

Ikani was available by SMS in case db1 needed vcenter related intervention in the middle of this upgrade. Rollback would be the use of the pg_dump backup which would be performed as the first step during this downtime.

* Our userbase was informed by @weasyldev and @weasyl's twitter account of the upgrade at -7 hours before upgrade time, and status.weasyl.com updated with the upcoming maintenance at that time. I outlined to gitter the length of the outage and a rough scale of the time the steps would take (10 minutes for backups, 20-25 minutes for the upgrade and vacuum processes). These times were taken from performing the same upgrade on db2, to determine estimates. I added additional time in case of other problems that might come up.
* Another tweet was sent to the userbase at -10 minutes as a reminder of the outage, and gitter.im's channel was also reminded. status.weasyl.com was put into a maintenance underway state at 11:00pm as the upgrade was started.

The upgrade itself involved the following steps:
* web1 and web2's workers were shut down. This brought up the 'Technical issue' error screen to users who attempted to access weasyl, and directed them to status.weasyl.com, which did show the maintenance page for the duration of this upgrade.
* Back up the old instance of postgresql with pg_dump. The dump file was copied to testdb for safekeeping. (17 minutes total).
* An ansible script was run, which shut down the old postgresql instance, ran 'pg_upgrade', then removed the 9.4 debian libraries once this was completed. New postgresql.conf files were installed. The new database was started up. (about 6 minutes)
* /usr/lib/postgresql/9.5/bin/vacuumdb --all --analyze-only was run, necessary after a pg_upgrade. (6 minutes)
* db1 was rebooted to assure that the database came up correctly, unattended. (It did so.)
[...]

PR #81, opened 2016-06-21
Use PostgreSQL 9.5’s ON CONFLICT to avoid exceptions and logged errors

charmander:
Addresses #51.

### DIFF
--- a/weasyl/controllers/api.py
+++ b/weasyl/controllers/api.py
@@ -480,11 +480,7 @@ class api_favorite_(api_base):
 
     @api_method
     def POST(self, content_type, content_id):
-        try:
-            favorite.insert(self.user_id, **{_CONTENT_IDS[content_type]: int(content_id)})
-        except WeasylError as we:
-            if we.value != 'favoriteRecordExists':
-                raise
+        favorite.insert(self.user_id, **{_CONTENT_IDS[content_type]: int(content_id)})
 
         return json.dumps({
             'success': True
--- a/weasyl/define.py
+++ b/weasyl/define.py
@@ -975,11 +975,11 @@ def common_view_content(userid, targetid, feature):
     else:
         viewer = get_address()
 
-    try:
-        engine.execute(
-            meta.tables['views'].insert()
-            .values(viewer=viewer, targetid=targetid, type=typeid))
-    except sa.exc.IntegrityError:
+    result = engine.execute(
+        'INSERT INTO views (viewer, targetid, type) VALUES (%(viewer)s, %(targetid)s, %(type)s) ON CONFLICT DO NOTHING',
+        viewer=viewer, targetid=targetid, type=typeid)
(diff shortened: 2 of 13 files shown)


weasyl/define.py
```diff
@@ -975,11 +975,11 @@ def common_view_content(userid, targetid, feature):
     else:
         viewer = get_address()
 
-    try:
-        engine.execute(
-            meta.tables['views'].insert()
-            .values(viewer=viewer, targetid=targetid, type=typeid))
-    except sa.exc.IntegrityError:
+    result = engine.execute(
```
hyena: Nice.

weasyl/controllers/api.py
```diff
@@ -480,11 +480,7 @@ class api_favorite_(api_base):
 
     @api_method
     def POST(self, content_type, content_id):
-        try:
-            favorite.insert(self.user_id, **{_CONTENT_IDS[content_type]: int(content_id)})
-        except WeasylError as we:
-            if we.value != 'favoriteRecordExists':
-                raise
+        favorite.insert(self.user_id, **{_CONTENT_IDS[content_type]: int(content_id)})
```
hyena: Christ that's ugly.

weasyl/searchtag.py
```diff
@@ -73,6 +73,21 @@ def select_list(map_table, targetids):
     return dict(list(db.execute(q)))
 
 
+@region.cache_on_arguments()
+def get_or_create(name):
+    name = d.get_search_tag(name)
+    tag = d.engine.execute(
+        'INSERT INTO searchtag (title) VALUES (%(name)s) ON CONFLICT (title) DO NOTHING RETURNING tagid',
```
hyena: Would an upsert style `ON CONFLICT (title) DO UPSERT` be a harmless noop here? It could save another query if so.

weasyl/test/test_collection.py
```diff
@@ -19,7 +19,22 @@ def offer(self):
         collection.offer(self.creator, self.s, self.collector)
 
     def count_collections(self, pending, rating=ratings.GENERAL.code):
-        return len(collection.select_manage(self.collector, rating, 10, pending))
+        return self.db.scalar(
```
hyena: This feels a little brittle. Is there nothing in the collection module that's suitable?

weasyl/searchtag.py
charmander: An update has to be specified; `ON CONFLICT (title) DO UPDATE SET tagid = tagid` would work, but feels like too much of a hack.

weasyl/test/test_collection.py
charmander: There was `collection.select_manage` =) It was only used for this set of tests, though, so I moved it here and took out the query string building. Note that it is still just a utility function and not a test.

PR #82, opened 2016-06-21
Add option to reload templates

charmander:
For your development convenience. Maybe there should be a `sass:watch` task at some point, too.


weasyl/define.py
```diff
@@ -204,21 +213,21 @@ def sql_string_series(target):
...
-    else:
-        template = template_name
-    if template is None:
+def compile(template_name):
+    """
+    Compiles a template file and returns the result.
+    """
+    template = _template_cache.get(template_name)
+
+    if template is None or reload_templates:
```
weykent: Considering template crap uses such a tiny amount of CPU it seems like you don't need anything more complex than this. Great! I don't know why this didn't get added earlier.

weasyl/define.py
```diff
@@ -1081,6 +1088,9 @@ def cdnify_url(url):
 
 
 def get_resource_path(resource):
+    if reload_templates:
```
weykent: Maybe 'reload templates' is too specific if it's going to be used here too?

charmander: Maybe, but I couldn’t come up with a better name. What describes the templates and the resources but not the non-template Python?

weykent: Assets? That might be too specific to non-templates. It looks like pyramid separately allows reloading of 'templates' and 'assets'. Is `reload_templates_and_assets` too verbose?

charmander: I can add a separate `reload_assets`.

PR #48, opened 2016-06-14
Indicate if profile user follows logged in user

dzamie:
(Suggested in Issue #8)
If a viewed user is following the current user, "Following You" will
appear to the left of "Follow" on their user page.
Added code to support an icon for "Following You", which could be placed
starting at (0, 260) of "./images/icons-20.png" (Current icons-20.png is
60x260, so an extra row would have to be added).

PR #83, opened 2016-06-22
Expose follower notification feature to the UI

kfkitsune:
Fully implements the follower notification feature initially proposed in Issue #8, and implemented on the backend in PR #48. Does so by adding an unobtrusive 'follows you' in the user information section of the user_tools.html template.

Example: https://i.sli.mg/KytdPL.png


charmander:
The location here is a bit awkward; someone can almost set their name to that. Alternative in the works.

PR #84, opened 2016-06-23
Use psycopg2cffi instead of psycopg2

charmander:
Compatible with PyPy and might resolve an elusive SystemError. `site.config.txt`s and `alembic.ini`s will have to be updated.

### DIFF
--- a/config/site.config.txt.example
+++ b/config/site.config.txt.example
@@ -17,7 +17,7 @@ time_endpoints = false
 rough_shutdowns = true
 
 [sqlalchemy]
-url = postgres:///weasyl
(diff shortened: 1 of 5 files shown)


weykent:
It is possible to use `psycopg2cffi.compat.register` instead of having to change the config. Does it matter if we never use the non-cffi version?

charmander:
That feels like a bit of a hack.

weykent:
I mean, yes, it is, but it saves having to find every use of `psycopg2` everywhere.

charmander:
I’m pretty sure this covers all of the instances in weasyl itself now.

PR #85, opened 2016-06-24
Remove 'cached=true' from templates

kfkitsune:
Finish removing cached parameters from templates; leaving the 'cached=true' in broke the template when attempting to view the template in question.

No further instances of ``cached = true`` exist in templates per ``git grep cached``.


charmander:
and this is why we format things properly

Issue #86, opened 2016-06-27
Show tags on results from /api/messages/submissions


taedixon:
[API documentation](https://projects.weasyl.com/weasylapi/#get--api-messages-submissions) indicates this route returns a list of Submission objects, which should include tags. However, no tags are actually returned. Either tags should be added to the response, or the API documentation should be updated to reflect the actual behaviour (preferably the former).

seanmcelroy:
I strongly recommend that tags be added!  For clients that use this who would like to build in filtering, it saves many repetitive calls to get those tags for search results

PR #87, opened 2016-06-27
Return tags with /api/messages/submissions API call

kfkitsune:
Returns a tags JSON item with a call to /api/messages/submissions, which the current version of the Weasyl HTTP API states should be happening anyway.

Example call to the API endpoint within the Weasyl development VM: https://paste.weasyldev.com/show/ZSm0yIoH10aADBCTKggv/

Proposed fix to #86.

### DIFF
--- a/weasyl/controllers/api.py
+++ b/weasyl/controllers/api.py
@@ -157,6 +157,8 @@ def tidy_submission(submission):
     if contype:
         submission['type'] = m.CONTYPE_PARSABLE_MAP[contype]
     submission['rating'] = ratings.CODE_TO_NAME[submission['rating']]
+    if 'tags' in submission:
+        submission['tags'] = submission['tags']
     submission['owner'] = submission.pop('username')
     submission['owner_login'] = d.get_sysname(submission['owner'])
     submission['media'] = submission.pop('sub_media')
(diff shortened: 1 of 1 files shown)


weasyl/controllers/api.py
```diff
@@ -157,6 +157,8 @@ def tidy_submission(submission):
     if contype:
         submission['type'] = m.CONTYPE_PARSABLE_MAP[contype]
     submission['rating'] = ratings.CODE_TO_NAME[submission['rating']]
+    if 'tags' in submission:
+        submission['tags'] = submission['tags']
```
charmander: This doesn’t look like a very useful line.

weasyl/controllers/api.py
```diff
@@ -451,6 +453,8 @@ def GET(self):
 
         ret = []
         for sub in submissions:
+            temp_sub = submission.select_view_api(self.user_id, sub['submitid'], anyway=True, increment_views=False)
```
charmander: Modify `message.select_submissions` to include tags instead?

kfkitsune:
For right now I am going to close this PR; I'll readily admit that while I _think_ I see what needs to be done (pivot from submission ID, to one or more searchmapsubmit, to searchtag), my SQL skills aren't strong enough here to adequately accomplish this. I also recognize that it is, as Charmander noted, a better idea to execute a single DB call in ``weasyl/message.py:select_submissions()`` as opposed to my initial theory of executing another query to retrieve more information than needed and only use a single returned item from that query.

Issue #52, opened 2016-06-14
Switch to pyca's bcrypt


weykent:
It's cffi and actively maintained: <https://github.com/pyca/bcrypt>

At some point we should probably figure out what other things should be switched to cffi equivalents for eventual pypy.

charmander:
Fixed by #88.

PR #88, opened 2016-06-30
Implements pyca/bcrypt

kfkitsune:
Changes out py-bcrypt for pyca/bcrypt==3.1.0. This version landed on PyPI a short while ago, and now includes the checkpw() function. Unicode still needed to be encoded to utf-8.

Fixes Issue #52.


charmander:
Included as f6230c731f662be1301bdc38f6ab9815c7441144.

PR #89, opened 2016-06-30
Partial schema cleanup

### LINKED ISSUE #68: Clean up the schema
Remove unused things:
- [x] `commission` table (never used, replaced by `commish*` tables)
- [x] `composition` table (part of long-removed Compose feature)
- [x] `contentview` table (appears to have been replaced by `views` table)
- [x] `ignorecontent` table (never used)
- [x] `loginaddress` table (never used)
- [x] `logininvite` table (we’re out of beta)
- [x] `disallow-others-tag-editing` flag in `profile.config` (never implemented)

Improve the representation of stuff:
- [ ] create a rating enumeration and use it instead of character codes and numbers
- [ ] separate character flag fields into columns
- [ ] use `TIMESTAMP WITH TIME ZONE` instead of Weasyl time (Unix time ± 18000)

Simplify miscellany:
- [x] Retire `disk_media`, extracting necessary information from `media` itself

Do something about that welcome table:
- [ ] no really it’s a huge problem

_This list is [incomplete](#wikipedia_joke); you can help by [expanding it](#new_comment_field)._

kfkitsune:
Partially resolves #68.

Of note, ``https://deploy.weasyldev.com/weasyl-latest-staff.sql.gz`` does not contain the below tables; therefore, to get the code to compile on the VM I needed to manually check to see if the tables even existed when the Alembic update() block is run; otherwise, ``vagrant provision`` would fail when it attempted to ``alembic upgrade head``.

I did manually downgrade to ``9afc9a45510c``, verified that the tables were recreated, and then upgraded, so the logic does work.

Items resolved are:

1. ``commission`` table (never used, replaced by ``commish*`` tables)
2. ``composition`` table (part of long-removed Compose feature)
3. ``contentview`` table (appears to have been replaced by ``views`` table)
4. ``ignorecontent`` table (never used)
5. ``loginaddress`` table (never used)
6. ``logininvite`` table (we’re out of beta)
7. ``disallow-others-tag-editing`` flag in ``profile.config`` (never implemented)


charmander:
Skimming this, it looks like you missed an instance of `loginaddress`.

kfkitsune:
And forgot that Charmander already has a branch for this. Closing this.

PR #93, opened 2016-07-05
Ads

charmander:
Basic support so we can see how they’re received by advertisers and browsers.


weasyl/errorcode.py
```diff
@@ -1,4 +1,4 @@
-# errorcode.py
+# coding=utf-8
```
kfkitsune: We should consider standardizing all files across the codebase to a single style when it comes to specifying file encoding. I am seeing the style you used appearing 1 time in master, ``# encoding: utf-8`` 2 times in master, and ``# -*- coding: utf-8 -*-`` 6 times in master.

On one of the tests I wrote, you suggested that file use the ``# encoding: utf-8`` variant, so perhaps standardize to that? It's the smallest of nits as the code is still functional, and I can get everything about the ads branch to work. If we are going for consistency, we should pick a style and use it.

weasyl/ads.py
```diff
@@ -0,0 +1,111 @@
...
+def _get_current_ads():
+    now = _serializable_datetime(datetime.now())
+    return [ad for ad in _get_all_ads() if ad["start"] <= now <= ad["end"]]
+
+
+def get_current_ads():
+    return [Ad(**ad) for ad in _get_current_ads()]
+
+
+def get_display_ads(userid, count):
```
kfkitsune: ``userid`` is seemingly unused, even though called through the templates. I can do the following and make the following display the ads as expected:

- $ display_ads = ads.get_display_ads(0, count=2)
- $ display_ads = ads.get_display_ads("foo", count=2)
- $ display_ads = ads.get_display_ads(userid="foofoo", count=2)

If it is supposed to serve a purpose (speculation: not seeing the user's own ads?), it is either not implemented yet, or it is not functioning as expected.

charmander: It’s intended to allow people to hide specific ads in the future. Not sure which features we want yet.

weasyl/ads.py
```diff
@@ -0,0 +1,112 @@
+import os
+import random
+import time
+
+from datetime import datetime
```
kfkitsune: PEP 8, Imports, https://www.python.org/dev/peps/pep-0008/#imports : 

> "Imports should be grouped in the following order: [...] 1) standard library imports [...]"

Shouldn't this be in the above block with ``os``, ``random``, and ``time``?

weasyl/controllers/ads.py
```diff
@@ -0,0 +1,50 @@
+from __future__ import absolute_import, division, unicode_literals
+
+from datetime import datetime
+from pyramid.httpexceptions import HTTPSeeOther
```
kfkitsune: PEP 8, Imports. "Related third-party imports" block starts here.

templates/ads/create.html
```diff
@@ -0,0 +1,53 @@
...
+            </label>
+
+            <label>
+                Link target:
+                <input type="url" name="target" value="" placeholder="https://example.com/" required />
+            </label>
+
+            <label>
+                End date (YYYY-MM-DD):
+                <input type="text" name="end" value="${(arrow.utcnow() + datetime.timedelta(days=30)).strftime('%Y-%m-%d')}" required />
```
taedixon: Is there a reason for not using type="date" here? Chrome at least has a built-in date picker that makes this a bit easier to set.

weasyl/ads.py
```diff
@@ -0,0 +1,112 @@
+import os
```
charmander: Yep, thanks!

weasyl/ads.py
charmander: Sure.

templates/ads/create.html
charmander: It was because of the `YYYY-MM-DD` hint – Chrome will display the date in a locale-dependent format – but I can add a script for that instead.

charmander:
Not happening in this version.

Issue #56, opened 2016-06-15
'SFW mode' doesn't filter out or indicate non-general journals


weykent:
Not even getting into comment or comment reply notifications, journal notifications for non-general-rated journals aren't filtered in the journal list. Clicking on the journal just gives the typical "rating too high" error page.

charmander:
Fixed by #94 as #95.

PR #94, opened 2016-07-05
Prevent non-SFW journals from displaying in notifications if SFW mode…

kfkitsune:
… is active

Resolves issue #56 by altering the statement passed to the database to only request SFW journals if user has SFW-mode active.


charmander:
Thanks for the fix! However, as the complete version (accounting for rating preference changes in general) is a one-liner, I'm just going to open #95. =)

kfkitsune:
Closing in favor of #95.

PR #96, opened 2016-07-05
Pyramid

hyena:
~~(Do not merge this yet; it is untested and incomplete. Opening this PR because it's a better way to track comments and change requests than #5.)~~ UPDATE: This is now ready for testing and review.

Ports weasyl from web.py to pyramid. web.py remains in order to render templates and to provide `storify()` for web forms (just in case any controllers involve subtlety) around that. The work is only partial work so far. The remaining work involves porting the controllers over, as well as the oauth2 logic, and the api calls.

Most of my thoughts are collected in the pyramid.md file, which is being used to track the work as it occurs to me. As the document attests, there'll still be work after functionality is there to make the system more idiomatic for pyramid.

Other good places to start reading the changes are middleware.py, define.py, and weasyl.controllers.routes.


weasyl/controllers/general.py
```diff
@@ -1,142 +1,136 @@
...
+        "title": "Streaming",
+    }
+    rating = define.get_rating(request.userid)
+    return Response(define.webpage(request.userid, 'etc/streaming.html',
+                                   [profile.select_streaming(request.userid, rating, 300, order_by="start_time desc")],
+                                   **extras))
+
+
+def site_update_(request):
+    # TODO: Handle a non-integer with something better than a generic error?
```
charmander: We can use a regular expression in the pattern and let it default to a 404, right?

hyena: Yep.

weasyl/api.py
```diff
@@ -9,7 +10,9 @@
 
 
 def is_api_user():
-    return 'HTTP_X_WEASYL_API_KEY' in web.ctx.env or 'HTTP_AUTHORIZATION' in web.ctx.env
+    # TODO: Pass in the request instead.
+    request = get_current_request()
+    return 'X_WEASYL_API_KEY' in request.headers or 'AUTHORIZATION' in request.headers
```
hyena: Note to self: Double check that this is correct.

weasyl/controllers/profile.py
```diff
@@ -1,526 +1,518 @@
...
+        raise WeasylError("userRecordMissing")
+    elif not request.userid and "h" in define.get_config(otherid):
+        return Response(define.errorpage(request.userid, errorcode.no_guest_access))
+
+    userprofile = profile.controller_baseselect_profile(otherid, images=True, viewer=request.userid)
+    has_fullname = userprofile['full_name'] is not None and userprofile['full_name'].strip() != ''
+    page_title = u"%s's collections" % (userprofile['full_name'] if has_fullname else userprofile['username'],)
+    page = define.common_page_start(request.userid, title=page_title)
+
+    url_format = "/c if name else ollections?userid={userid}&%s".format(userid=userprofile['userid'])
```
charmander: This doesn’t look right somehow.

hyena: Thanks for catching that. That's a bad trackpad driver doing a lot of pasting when I scroll in pycharm.

weasyl/define.py
```diff
@@ -110,16 +109,14 @@ def userid_request_property(request):
...
+            raise HTTPUnauthorized(headers={'WWW-Authenticate': 'Weasyl-API-Key realm="Weasyl"'})
         return userid
 
     elif authorization:
         from weasyl.oauth2 import get_userid_from_authorization
         userid = get_userid_from_authorization()
         if not userid:
-            request.response.headers['WWW-Authenticate'] = 'Bearer realm="Weasyl" error="invalid_token"'
-            raise HTTPUnauthorized()
+            raise HTTPUnauthorized({'WWW-Authenticate': 'Bearer realm="Weasyl" error="invalid_token"'})
```
charmander: Specify `headers=` here too?

weasyl/middleware.py
```diff
@@ -101,25 +101,24 @@ def session_tween(request):
...
+        # Register a response callback to clear and set the session cookies before returning.
+        # Note that this requires that exceptions are handled properly by our exception view.
+        def callback(request, response):
             if sess_obj.save:
                 session.begin()
                 if sess_obj.create:
                     session.add(sess_obj)
-                    resp.set_cookie('WZL', sess_obj.sessionid, max_age=60 * 60 * 24 * 365,
+                    response.set_cookie('WZL', sess_obj.sessionid, max_age=60 * 60 * 24 * 365,
                                     secure=request.scheme == 'https', httponly=True)
```
charmander: This is misaligned now. I wish we would just indent things.

weasyl/middleware.py
```diff
@@ -369,7 +369,7 @@ def ravenCaptureArguments(self, level=None, **extra):
             'data': data,
             'extra': dict(
                 extra,
-                session=request.get('weasyl_session'),
+                session=request.weasyl_session if hasattr(request, 'weasyl_session') else None
```
charmander:
```python
session=getattr(request, 'weasyl_session', None),
```

hyena: Right. Sorry, jetlag coding.

PR #97, opened 2016-07-06
Improve Recently Popular performance without index

charmander:
PostgreSQL insists on picking the new unixtime index for other favorites operations, making them run slowly. We could do a bunch of manual tuning, or do this instead, because it’s more efficient.

### DIFF
--- a/weasyl/submission.py
+++ b/weasyl/submission.py
@@ -1054,6 +1054,11 @@ def select_recently_popular():
     max_days = int(d.config_read_setting("popular_max_age_days", "21"))
 
     query = d.engine.execute("""
+        WITH t AS (
+            SELECT submitid AS cutoff FROM submission
+            WHERE unixtime > EXTRACT(EPOCH FROM now() - %(max_days)s * INTERVAL '1 day')::INTEGER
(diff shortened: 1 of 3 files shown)


weasyl/submission.py
```diff
@@ -1072,7 +1077,7 @@ def select_recently_popular():
             LEFT JOIN favorite ON favorite.type = 's' AND submission.submitid = favorite.targetid
         WHERE
             submission.unixtime > EXTRACT(EPOCH FROM now() - %(max_days)s * INTERVAL '1 day')::INTEGER AND
```
hyena: Could this be
WHERE
    submission.submitid >= (SELECT cutoff FROM t)?

That would also simplify the favorite where clause.

charmander: Yes, it could be; thanks! However, it seems this isn’t necessary at all; PostgreSQL can do the join properly now. I don’t know if this is because we ran an `ANALYZE` on the favorite table, or because SQLAlchemy’s generated query was just that bad.

hyena: Probably the latter.

PR #98, opened 2016-07-06
Index media by sha256

charmander:
Stops `MediaItem.fetch_or_create()` from scanning the entirety of the media table, a process that currently slows every upload down by about a second.


skylerbunny:
WAD on testing.

...
         ->  Index Scan using ind_media_sha256 on media  (cost=0.56..8.57 rows=1 width=116) (actual time=0.021..0.021 rows=0 loops=1)
...
 Planning time: 0.411 ms
 Execution time: 0.062 ms

PR #109, opened 2016-07-17
Return tags with /api/messages/submissions call

kfkitsune:
The API says tags should already be returned for the ``/api/messages/submissions`` endpoint; so return them. Proposed fix for #86.

Example output, using three submissions, one character. https://paste.weasyldev.com/show/Bl4xWt1qQ9egqlFJkPYO/

### DIFF
--- a/weasyl/message.py
+++ b/weasyl/message.py
@@ -3,7 +3,6 @@ from itertools import chain
 from weasyl import character
 from weasyl import define as d
 from weasyl import media
-from weasyl import searchtag
 
 
 notification_clusters = {
--- a/weasyl/message.py
+++ b/weasyl/message.py
@@ -117,7 +117,7 @@ def select_submissions(userid, limit, backtime=None, nexttime=None):
                 INNER JOIN character ch ON we.targetid = ch.charid
                 INNER JOIN profile pr ON ch.userid = pr.userid
                 LEFT JOIN LATERAL (
-                    SELECT c.charid AS id, string_agg(tags.title, ',') AS tags
+                    SELECT c.charid AS id, array_agg(tags.title) AS tags
                     FROM character as c
                     LEFT JOIN searchmapchar AS cmap ON c.charid = cmap.targetid
                     JOIN searchtag AS tags ON cmap.tagid = tags.tagid
@@ -143,7 +143,7 @@ def select_submissions(userid, limit, backtime=None, nexttime=None):
                 INNER JOIN submission su ON we.targetid = su.submitid
                 INNER JOIN profile pr ON we.otherid = pr.userid
                 LEFT JOIN LATERAL (
-                    SELECT s.submitid AS id, string_agg(tags.title, ',') AS tags
+                    SELECT s.submitid AS id, array_agg(tags.title) AS tags
                     FROM submission AS s
                     LEFT JOIN searchmapsubmit AS smap ON s.submitid = smap.targetid
                     JOIN searchtag AS tags ON smap.tagid = tags.tagid
@@ -169,7 +169,7 @@ def select_submissions(userid, limit, backtime=None, nexttime=None):
                 INNER JOIN submission su ON we.targetid = su.submitid
                 INNER JOIN profile pr ON su.userid = pr.userid
                 LEFT JOIN LATERAL (
-                    SELECT s.submitid AS id, string_agg(tags.title, ',') AS tags
+                    SELECT s.submitid AS id, array_agg(tags.title) AS tags
                     FROM submission AS s
                     LEFT JOIN searchmapsubmit AS smap ON s.submitid = smap.targetid
                     JOIN searchtag AS tags ON smap.tagid = tags.tagid
@@ -206,7 +206,7 @@ def select_submissions(userid, limit, backtime=None, nexttime=None):
         "userid": i.userid,
         "username": i.username,
         "subtype": i.subtype,
-        "tags": (i.tags).split(','),
+        "tags": i.tags,
         "sub_media": _fake_media_items(i),
     } for i in query]
 
--- a/weasyl/message.py
+++ b/weasyl/message.py
@@ -3,6 +3,7 @@ from itertools import chain
 from weasyl import character
 from weasyl import define as d
 from weasyl import media
+from weasyl import searchtag as st
 
 
 notification_clusters = {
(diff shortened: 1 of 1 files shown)


weasyl/message.py
```diff
@@ -3,6 +3,7 @@
 from weasyl import character
 from weasyl import define as d
 from weasyl import media
+from weasyl import searchtag as st
```
hyena: Don't rename this, imo, since it's non-obvious what 'st' stands for and it's only used in a few places.

weasyl/message.py
```diff
@@ -168,18 +169,25 @@ def select_submissions(userid, limit, backtime=None, nexttime=None):
...
-        "subtype": i.subtype,
-        "sub_media": _fake_media_items(i),
-    } for i in query]
+    results = []
+    for i in query:
+        if i.contype != _CONTYPE_CHAR:
+            tags = st.select(submitid=i.id)
+        else:
+            tags = st.select(charid=i.id)
+        results.append({
```
hyena: Why not do a list comprehension for this?

hyena:
This probably works. I'm a little concerned about the performance. Are there other instances where we do a separate `searchtag.select()` for items in a loop? (As opposed to retrieving them along with the submissions/characters with a join).

kfkitsune:
> Are there other instances where we do a separate searchtag.select() for items in a loop?

I checked the codebase, and no, all other calls to ``searchtag.select()`` are one-offs, though there is a peculiar little block of code in ``index.py::recent_submissions()`` [lines 23 to 44](https://github.com/Weasyl/weasyl/blob/97414b45bb5de0b24fe4c7dcd68f0d95bbeaf140/weasyl/index.py#L23), which almost appears to do something almost, but not quite similar; at least it feels or looks like it is operating on a loop against tables to get tags out (though it's not using ``select()`` for it).

> (As opposed to retrieving them along with the submissions/characters with a join).

I can certainly try this, as I _think_ I know the basics of what needs to be done, but I have a bit of an issue. When I tried this previously, my attempt at getting the tags didn't quite work, and ended up returning a new row entirely (or it looked like it) for each of the three tags I was setting. So while it feels like I was close, I don't think I was right where I was needing to be. The parts of the query which were modified can be found [here](https://paste.weasyldev.com/show/heTSQfcs4dErl8CoajcR/); the bits which aren't captured in the paste are the same as they are in the file in master.

If you can give me a nudge in the right direction, I can certainly take another stab at extracting the array of tags within the main SQL query.

hyena:
My sql isn't the strongest, but postgresql has a handy `array_agg()` function:

```
weasyl=# select s.submitid, array_agg(tags.title) FROM submission as s LEFT JOIN searchmapsubmit as map on s.submitid=map.targetid JOIN searchtag as tags ON map.tagid=tags.tagid GROUP BY s.submitid LIMIT 10;
 submitid |                                array_agg                                 
----------+--------------------------------------------------------------------------
     2031 | {wesley,weasel,beret,ermine,vest,weasyl,male,anthro}
     2032 | {wesley,weasel,beret,ermine,vest,weasyl,male,sleeping,anthro}
     2072 | {wesley,weasel,writing,beret,vest,weasyl,male,anthro}
     2104 | {space,wallpaper,space_core,portal_2,planet,taw}
     2107 | {logo,weasyl,awesome,9000_hours_in_ms_paint}
     2108 | {wesley,weasel,mp3_player,beanbag_chair,beret,ermine,vest,weasyl,anthro}
     2117 | {in,a,tiger,tie}
     2131 | {wesley,weasel,weasyl,beret,ermine,vest,male,anthro}
     2157 | {angel,dark_pit,kid_icarus,pit,wings,male}
     2249 | {jumping,beret,ermine,vest,happy,weasyl,weasel,wesley,male,anthro}
(10 rows)
```

I don't know how that maps onto the output of sqlalchemy, but it might be straightforward.

kfkitsune:
Example output, using one of each submission type that the query returns (collection, character, submission): https://paste.weasyldev.com/show/fzoalbspXCkjxAcW2693/

weasyl/message.py
```diff
@@ -110,10 +111,19 @@ def select_submissions(userid, limit, backtime=None, nexttime=None):
                 pr.username,
                 ch.settings,
                 we.welcomeid,
-                0 AS subtype
+                0 AS subtype,
+                agg.tags
```
charmander: Use `array_agg` as @hyena suggested, please.

weasyl/message.py
```diff
@@ -110,10 +111,19 @@ def select_submissions(userid, limit, backtime=None, nexttime=None):
...
             FROM welcome we
                 INNER JOIN character ch ON we.targetid = ch.charid
                 INNER JOIN profile pr ON ch.userid = pr.userid
+                LEFT JOIN LATERAL (
```
charmander: Why does this need to be a `LATERAL` join?

kfkitsune: Based on my understanding of the keyword, it permits referencing already referenced columns; without it the subquery would be evaluated independently (see [LATERAL Subqueries](https://www.postgresql.org/docs/9.5/static/queries-table-expressions.html#QUERIES-LATERAL)). Based on how I am currently understanding the tables, the link would be ``character|submission id 1:N searchmap(char|submit) 1:1 searchtag``; the ID of the submission item needs to be compared against the ``targetid`` in the searchmap for the item being examined at that point in the overall query, hence the lateral which allows--in this instance--``ch.charid`` to be referenced and matched to any ``searchmapchar`` rows which have that ID in ``cmap.targetid``.

If there's an alternate method, let me know and I'll give it a shot; this method was the first that I found which worked, and hence why I selected it.

Not including ``LATERAL`` results in the following (stripping ``LATERAL`` without changing the subquery).

```
2016-07-19T01:34:51+0000 [stderr#error] ProgrammingError: (psycopg2cffi._impl.exceptions.ProgrammingError) invalid reference to FROM-clause entry for table "su"
2016-07-19T01:34:51+0000 [stderr#error] LINE 75:                     WHERE smap.targetid = su.submitid
2016-07-19T01:34:51+0000 [stderr#error]                                                    ^
2016-07-19T01:34:51+0000 [stderr#error] HINT:  There is an entry for table "su", but it cannot be referenced from this part of the query.
```

weasyl/message.py
kfkitsune: Figured this one out; was having issues initially, but I got it to work. I'll be synchronizing it in a bit.

weasyl/message.py
```diff
@@ -110,13 +111,20 @@ def select_submissions(userid, limit, backtime=None, nexttime=None):
...
                 ch.settings,
                 we.welcomeid,
-                0 AS subtype
+                0 AS subtype,
+                array_agg(tags.title) as tags
             FROM welcome we
                 INNER JOIN character ch ON we.targetid = ch.charid
                 INNER JOIN profile pr ON ch.userid = pr.userid
+                LEFT JOIN searchmapchar AS smc ON ch.charid = smc.targetid
+                JOIN searchtag AS tags USING (tagid)
```
charmander: I think this should be a `LEFT JOIN` too.

weasyl/message.py
```diff
@@ -110,13 +111,20 @@ def select_submissions(userid, limit, backtime=None, nexttime=None):
                 pr.username,
                 ch.settings,
                 we.welcomeid,
-                0 AS subtype
+                0 AS subtype,
+                array_agg(tags.title) as tags
```
charmander: Uppercase `AS` for consistency.

PR #115, opened 2016-08-05
Avoid checking tokens in api_method decorator

charmander:
It makes no sense and bypasses `login_required`, causing [errors][1].

  [1]: https://sentry.weasyldev.com/weasyl/weasyl/issues/1416/

### DIFF
--- a/weasyl/controllers/api.py
+++ b/weasyl/controllers/api.py
@@ -45,11 +45,6 @@ _CONTENT_IDS = {
 
 def api_method(f):
     def wrapper(self, *a, **kw):
-        form = web.input(token="")
-
-        if not api.is_api_user() and form.token != d.get_token():
-            self.user_id = 0
-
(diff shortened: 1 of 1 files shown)


hyena:
One of the consequences of removing this if-condition is that cookies can be used to identify a user in cases where they weren't before. Is there any way that could be a problem?

Do we know why this was something we tried to avoid in the first place?

charmander:
> One of the consequences of removing this if-condition is that cookies can be used to identify a user in cases where they weren't before. Is there any way that could be a problem?

The API shouldn’t be usable in a browser except by `www.weasyl.com`; as far as I can tell, we’re not allowing cross-origin access to it.

It was introduced by a commit whose hash ends with d85f8cf47\* because I apparently forgot about the same-origin policy in August 2015.

\* Not providing a prefix because GitHub still has access to this commit and you can walk the entire history of weasyl-old from there. Nice.

Issue #121, opened 2016-08-11
Remove "Non-original audio upload" report type


Hendikins:
(Previously GitLab issue 27)

Per discussion in -admin, this report type is little used and considered redundant.

*   Report type to be removed
*  Existing reports of this type to be changed to "Tracing or plagiarism" (type 2020)

PR #138, opened 2016-08-11
Remove “Non-original audio upload” report type

charmander:
Addresses #121.


kfkitsune:
> What if someone wants to go back several steps and this is one of them?
Personally I'd say write text to the console, if anything, such as, "The Alembic revision <revID> modified rows. Reversion of changes not possible, however downgrading revision level as requested."

I find myself agreeing with Hyena, here, to be quite honest.

charmander:
@hyena They can update the version identifier manually.

Issue #141, opened 2016-08-11
Director control panel


Hendikins:
Per policy decision in -directors - certain tools to require higher level of authorisation than administrator.

Initial items for director control panel:
#99 (e-mail domain blacklist)
#140 (global tag addition blacklist)
Impersonate user

PR #111, opened 2016-07-18
Add email blacklist to block email domains during registration

kfkitsune:
~~- [ ] Depends on #143 being merged (function provides a higher-level permission than admin).~~
~~- [ ] When #143 is merged, re-code this code to move it to the director's panel.~~
- [x] To Pyramid.
- [x] Add the director control panel to this PR.

---

For Issue #99.

Adds a director-only page to permit blocking specified email providers during registration. Domain names provided to the page are stored in the database, and upon a new user attempting to sign up, the domain of the email is checked against the blacklist; if present, an error page is displayed.

Removal of entries from the blacklist functions via a checkbox system (not unlike submission removal), providing for ease of maintenance of outdated or incorrectly added entries.

PR #143, opened 2016-08-11
Add director control panel

kfkitsune:
- [x] To Pyramid.

---
For #141.

Extends ``weasyl/controllers/base.py`` to permit checking for users in the ``staff.DIRECTORS`` frozenset.

Shifts the Impersonate User function to the newly added Director Control Panel, removing it from the Admin Control Panel.


charmander:
Impersonation is going to be removed, so once #159 is merged, could you add the director control panel in #111 instead, please?

Issue #139, opened 2016-08-11
Move /modcontrol/finduser to /admincontrol


Hendikins:
This tool can be used to search for/obtain personal information (e-mail addresses, IPs if #128 implemented through this tool).

Per discussion in -directors, move to admin tools for following reasons:
* This tool is not required for any moderator-level actions
* Standardise access level required to access personal information

PR #144, opened 2016-08-12
Move /modcontrol/finduser to /admincontrol/finduser

kfkitsune:
For #139.

Moves /modcontrol/finduser to /admincontrol/finduser.


charmander:
There’s lots to clean up in the template, but these changes don’t introduce anything bad,

PR #148, opened 2016-08-26
Image cleanup

charmander:
- [ ] Add tests, maybe

### DIFF
--- a/weasyl/banner.py
+++ b/weasyl/banner.py
@@ -1,5 +1,5 @@
-from libweasyl import ratings
 from libweasyl import media
+from libweasyl import ratings
 from libweasyl.models.media import UserMediaLink
 
 
--- a/weasyl/banner.py
+++ b/weasyl/banner.py
@@ -1,8 +1,6 @@
-# banner.py
-
 from libweasyl import ratings
-
-from weasyl import media, orm
+from libweasyl import media
+from libweasyl.models.media import UserMediaLink
 
 
 def upload(userid, filedata):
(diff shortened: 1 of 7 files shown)


weasyl/banner.py
```diff
@@ -1,8 +1,6 @@
-# banner.py
-
 from libweasyl import ratings
-
-from weasyl import media, orm
+from libweasyl import media
```
hyena: nit: Lexicographic ordering

weasyl/image.py
```diff
@@ -50,74 +14,24 @@ def image_setting(im):
...
-
-    # if there's no need to resize, in-place resize is a no-op. otherwise copy
-    # the source to the destination.
-    elif not in_place:
-        files.copy(filename, destination)
-
-
-def resize_image(im, width, height):
-    return correct_image_and_call(_resize, im, width, height) or im
+        return im.original_format in ['JPEG', 'PNG', 'GIF']
```
hyena: This part could be made a little stronger. afaict this is only used in the case of thumbnails for characters (ugh). Maybe we should not only ensure that the original_format is valid but also that the format matches the file name. Doesn't hurt to sanity check that (maybe even throw an exception if we're in a state that's permitted a file to be named something that isn't its correct type).

hyena:
Looks good and a simpler world (+45 −198). Thanks for doing this.

+1.

weasyl/image.py
charmander: I’d like to err on the side of maintaining existing behaviour for this pull request, and just remove all the character special casing in a future one.

PR #150, opened 2016-08-29
Remove leftover `self`s from controller helpers

charmander:
Introduced in d473003b9be13ce0687fc78c39321a1832d3413d.


hyena:
Oops.

Thanks for fixing my mess.

PR #151, opened 2016-08-29
Miscellaneous bugfixes

charmander:
These popped up in Sentry, but only once, because nobody uses collections.


hyena:
Oops. Thanks for fixing my mess.

PR #152, opened 2016-08-29
Fix iteritems call

charmander:
https://sentry.weasyldev.com/weasyl/weasyl/issues/1433/

@hyena, does this need to be inserted somewhere else?


hyena:
I'm pretty sure this is was just a typo.

PR #154, opened 2016-08-30
Encode OAuth headers

charmander:
Essentially, this doesn’t work at all right now.

### DIFF
--- a/weasyl/oauth2.py
+++ b/weasyl/oauth2.py
@@ -6,12 +6,20 @@ from oauthlib.oauth2 import FatalClientError, OAuth2Error
 from libweasyl.oauth import get_consumers_for_user, revoke_consumers_for_user, server
 from weasyl.controllers.decorators import disallow_api, token_checked
 from weasyl import define as d
-from weasyl import errorcode, http, login, media, orm
+from weasyl import errorcode, login, media, orm
+
+
+class OAuthResponse(Response):
+    def __init__(self, headers, body, status):
+        super(OAuthResponse, self).__init__(
+            body=body,
+            status_code=status,
+            headers={k.encode('utf-8'): v.encode('utf-8') for k, v in headers.iteritems()},
+        )
 
 
 def extract_params(request):
-    headers = http.get_headers(request.environ)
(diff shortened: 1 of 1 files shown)


weasyl/oauth2.py
```diff
@@ -66,21 +74,15 @@ def authorize_post_(request):
                                     form.username, form.password, bool(form.remember_me),
                                     bool(form.not_me)))
     credentials['userid'] = userid
-    headers, body, status = server.create_authorization_response(
+    response = server.create_authorization_response(
```
hyena: I'd prefer to reserve the name `response` for Pyramid response objects only. Everywhere else in the code-base that's what it refers to, so avoiding breaking that here was a part of least surprise.

weasyl/oauth2.py
```diff
@@ -66,21 +74,15 @@ def authorize_post_(request):
...
 
 
 @disallow_api
 def token_(request):
-    headers, body, status = server.create_token_response(*extract_params())
-    for k, v in headers.iteritems():
-        request.set_header_on_response(k, v)
-    request.set_status_on_response("%s Status" % (status,))
-    return Response(body)
+    response = server.create_token_response(*extract_params(request))
```
hyena: Here as well.

weasyl/oauth2.py
```diff
@@ -6,12 +6,20 @@
...
+            body=body,
+            status_code=status,
+            headers={k.encode('utf-8'): v.encode('utf-8') for k, v in headers.iteritems()},
+        )
 
 
 def extract_params(request):
-    headers = http.get_headers(request.environ)
-    return request.path, request.method, request.params, headers
+    return request.path, request.method, request.params, request.headers
```
hyena: Should https://github.com/Weasyl/weasyl/blob/9dea21db3c8ddb22dbf804366834df9d2feb7f5a/weasyl/middleware.py#L350 change as well?

hyena:
Inheritance, eh. Alright.

weasyl/oauth2.py
charmander: Probably. For another pull request.

Issue #158, opened 2016-08-30
Remove Impersonate User feature


ikanimew:
Impersonate User was originally created for finding bugs, and is honestly no longer needed. Any mod function that would need Impersonate should just have a control in the Mod, Admin or Director panels, and we can approach those as they come up.

I believe there was originally discussion about moving this feature to the new Director panel, but we can just remove it entirely.

ikanimew:
Adding that #147 does block this.

hyena:
Pushed in #159.

PR #159, opened 2016-08-30
Remove “impersonate user” feature

charmander:
Now nothing is stored in `additional_data`. SFW mode probably should be, though.

Addresses #158.


weasyl/login.py
```diff
@@ -33,11 +33,6 @@ def signout(request):
     sess = request.weasyl_session
     # unset SFW-mode cookie on logout
     request.delete_cookie_on_response("sfwmode")
-    if sess.additional_data.get('user-stack'):
```
hyena: Nice. This will probably let us simplify other stuff.

PR #162, opened 2016-09-01
Cleanup

charmander:
Miscellany.

### DIFF
--- a/weasyl/api.py
+++ b/weasyl/api.py
@@ -1,5 +1,3 @@
-# api.py
-
 from __future__ import absolute_import
 
 from pyramid.threadlocal import get_current_request
--- a/weasyl/api.py
+++ b/weasyl/api.py
@@ -1,5 +1,7 @@
 # api.py
 
+from __future__ import absolute_import
+
 from pyramid.threadlocal import get_current_request
 
 from libweasyl.models.meta import Base
--- a/weasyl/character.py
+++ b/weasyl/character.py
@@ -1,33 +1,30 @@
-# character.py
+from __future__ import absolute_import
 
 import arrow
 import re
-
-from error import PostgresError, WeasylError
-import macro
-import define
-
-import image
-import comment
-import profile
-import welcome
-import blocktag
-import searchtag
-import thumbnail
-import frienduser
-import ignoreuser
-
-import files
-import report
-import favorite
 
 from libweasyl import ratings
 from libweasyl import staff
 from libweasyl import text
 
 from weasyl import api
+from weasyl import blocktag
+from weasyl import comment
+from weasyl import define
+from weasyl import favorite
+from weasyl import files
+from weasyl import frienduser
+from weasyl import ignoreuser
+from weasyl import image
+from weasyl import macro
 from weasyl import macro as m
 from weasyl import media
+from weasyl import profile
+from weasyl import report
+from weasyl import searchtag
+from weasyl import thumbnail
+from weasyl import welcome
+from weasyl.error import PostgresError, WeasylError
 
 
 _MEGABYTE = 1048576
--- a/weasyl/character.py
+++ b/weasyl/character.py
@@ -93,7 +93,7 @@ def create(userid, character, friends, tags, thumbfile, submitfile):
     ch = define.meta.tables["character"]
 
     try:
-        charid = define.engine.execute(ch.insert().returning(ch.c.charid), {
+        charid = define.engine.scalar(ch.insert().returning(ch.c.charid), {
             "userid": userid,
             "unixtime": arrow.now(),
             "char_name": character.char_name,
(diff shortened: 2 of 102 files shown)


weasyl/controllers/user.py
```diff
@@ -109,9 +111,9 @@ def signup_post_(request):
     form = request.web_input(
         username="", password="", passcheck="", email="", emailcheck="",
         day="", month="", year="", recaptcha_challenge_field="",
-        g_recaptcha_response=request.web_input("g-recaptcha-response"))
+        g_recaptcha_response="")
```
kfkitsune: This breaks the login form, as an FYI. https://paste.weasyldev.com/show/nPaojKrYSILAx5JawNwS/

I don't have time enough at the moment for a deep-dive fix on this, but the ``request.web_input(...)`` was in there because of compile issues, since the reCAPTCHA code uses dashes, and dashes don't/didn't play well with the code when I initially implemented the v2 code; simply put, there's no field on the form that is named ``g_recaptcha_response``. It will be named ``g-recaptcha-response`` when the JavaScript code loads.

weasyl/controllers/user.py
```diff
@@ -109,9 +111,9 @@ def signup_post_(request):
...
 
-    if not define.captcha_verify(form):
+    if not define.captcha_verify(form.g_captcha_response):
```
kfkitsune: ``g_recaptcha_response``, or if you can get the dashes to work, ``g-recaptcha-response``. There's no form field named ``g_captcha_response``, either on the form itself or inserted by the JavaScript reCAPTCHA code.

weasyl/controllers/user.py
charmander: Yeah, that wasn’t the right way to fix that. The kwargs passed to `web.input()` are defaults for fields with the same name. If a field is required, you can just leave it off, or pass it as an unnamed argument.

```python
form = request.web_input('g-recaptcha-response', …)
```

and access as `form['g-recaptcha-response']`.

weasyl/character.py
```diff
@@ -1,33 +1,30 @@
...
+from weasyl import blocktag
+from weasyl import comment
+from weasyl import define
+from weasyl import favorite
+from weasyl import files
+from weasyl import frienduser
+from weasyl import ignoreuser
+from weasyl import image
+from weasyl import macro
 from weasyl import macro as m
```
hyena: Looks like the same as the previous import. I'd prefer just one.

weasyl/define.py
```diff
@@ -349,7 +350,7 @@ def get_userid():
 
 
 def get_token():
-    import api
+    from weasyl import api
```
hyena: Ick.

weasyl/resetpassword.py
```diff
@@ -1,6 +1,8 @@
 # resetpassword.py
 # -*- coding: utf-8 -*-
```
kfkitsune: ``# encoding: utf-8``

Also, ``weasyl/test/test_define.py`` has the same change needed.

weasyl/api.py
```diff
@@ -1,5 +1,7 @@
 # api.py
```
kfkitsune: If we are removing the ``# <file>`` comments, the following files were missed (looking through just the changed file list here in this PR):

api, avatar, banner, content, define, emailer, embed, errorcode, files, folder, ignoreuser, login, note, premiumpurchase, resetpassword, search, searchtag, shout, welcome

weasyl/followuser.py
```diff
@@ -152,6 +149,7 @@ def insert(userid, otherid):
         'ON CONFLICT DO NOTHING',
         user=userid, other=otherid, settings=WatchSettings.from_code(d.get_config(userid)).to_code())
 
+    from weasyl import welcome
```
kfkitsune: PEP 8 seems to apply here, and at L165? I know it technically _works_, but PEP seems to say to keep it up top?

> Imports are always put at the top of the file, just after any module comments and docstrings, and before module globals and constants

https://www.python.org/dev/peps/pep-0008/#imports

weasyl/resetpassword.py
charmander: I guess I can cover that in this cleanup.

weasyl/api.py
charmander: Just removed in files I changed manually.

weasyl/followuser.py
charmander: Mutually-dependent modules.

weasyl/polecat.py
```diff
@@ -1,5 +1,7 @@
 """Polecats are twisted weasyls.
 
+from __future__ import absolute_import
```
kfkitsune: This will never get executed. Docstring begins at L1, ends at L6, as of this commit.

weasyl/read_staff_yaml.py
```diff
@@ -1,4 +1,6 @@
 """
+from __future__ import absolute_import
```
kfkitsune: This will never get executed. Docstring begins at L1, and ends at L36 of this commit.

charmander: Oh, nice catch. Apparently my rewriter is buggy.

charmander:
Thanks for the review!

Issue #134, opened 2016-08-11
Add note received from user to staff notes


Hendikins:
(Originally filed as Redmine issue 458)

From time to time we receive correspondence from users that needs to be recorded on their staff notes.

Adding " Copy this note to the users' staff notes" to note page will enable us to do this in the same manner as recording notes sent to users.

PR #169, opened 2016-09-23
Add ability to copy received note to staff notes

kfkitsune:
For #134.

Adds a button to the note page of a received note, which permits moderators to directly copy the contents of the note to the Staff Notes of the user who sent the note.

I opted to put the controller and code in the ``moderation`` section since it is only for moderators, however if there is a desire for it, I can shift it under the ``interaction`` controller and ``note.py``; as it is currently written, though, no additional imports are needed.

### DIFF
--- a/templates/note/message_view.html
+++ b/templates/note/message_view.html
@@ -34,7 +34,7 @@ $:{RENDER("common/stage_title.html", ["View Private Message", "Notes", "/notes?f
 
     $if myself['is_mod']:
       <form action="/modcontrol/copynotetostaffnotes" method="post" class="form clear">
-        <button class="button positive" style="float: right;">Copy Note to Staff Notes</button>
+        <button class="button" style="float: right;">Copy Note to Staff Notes</button>
         <input type="hidden" name="noteid" value="${query['noteid']}" />
         $:{CSRF()}
       </form>
--- a/templates/note/message_view.html
+++ b/templates/note/message_view.html
@@ -32,6 +32,17 @@ $:{RENDER("common/stage_title.html", ["View Private Message", "Notes", "/notes?f
 
     <div class="formatted-content">$:{MARKDOWN(query['content'])}</div>
 
+    $if myself['is_mod']:
+      <form name="copynotetostaffnotes" action="/modcontrol/copynotetostaffnotes" method="post" class="form clear">
+        <button type="submit" class="button positive" style="float: right;">Copy Note to Staff Notes</button>
+        <input type="hidden" name="notecontent" value="$:{MARKDOWN(query['content'])}" />
+        <input type="hidden" name="notesubject" value="${query['title']}" />
+        <input type="hidden" name="targetuserid" value="${query['senderid']}" />
+        <input type="hidden" name="targetusername" value="${LOGIN(query['sendername'])}" />
+        <input type="hidden" name="notedatetime" value="${(arrow.get(query['unixtime']).format('YYYY-MM-DD HH:mm:ss ZZ'))}" />
+        $:{CSRF()}
+      </form>
+
     $if not query['mine']:
       <form name="notescompose" action="/notes/compose" method="post" class="form clear">
         $:{CSRF()}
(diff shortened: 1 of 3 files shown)


templates/note/message_view.html
```diff
@@ -32,6 +32,17 @@
 
     <div class="formatted-content">$:{MARKDOWN(query['content'])}</div>
 
+    $if myself['is_mod']:
+      <form name="copynotetostaffnotes" action="/modcontrol/copynotetostaffnotes" method="post" class="form clear">
+        <button type="submit" class="button positive" style="float: right;">Copy Note to Staff Notes</button>
+        <input type="hidden" name="notecontent" value="$:{MARKDOWN(query['content'])}" />
```
charmander: I think this should use the note id instead of submitting all the information through a form. If that weren’t the case, though, you still wouldn’t want to use `$:` here, or pass the content through the Markdown filter for that matter.

weasyl/controllers/moderation.py
```diff
@@ -190,3 +190,21 @@ def modcontrol_edituserconfig_(request):
...
     raise HTTPSeeOther("/modcontrol")
+
+
+@moderator_only
+@token_checked
+def modcontrol_copynotetostaffnotes_post_(request):
+    form = request.web_input(targetuserid=None, targetusername=None, notecontent=None,
+                             notedatetime=None, notesubject=None)
+
+    staffnotetitle = ''.join(("Received note from ", form.targetusername, ", dated ", form.notedatetime,
```
charmander: You can use `%`-formatting to make this clearer.

weasyl/controllers/moderation.py
```diff
@@ -190,3 +190,21 @@ def modcontrol_edituserconfig_(request):
...
+                              ', and the subject was: "', form.notesubject, '".'))
+
+    moderation.note_about(
+        userid=request.userid,
+        target_user=form.targetuserid,
+        title=staffnotetitle,
+        message=form.notecontent,
+    )
+    raise HTTPSeeOther(''.join(("/staffnotes/", form.targetusername)))
```
charmander:
```python
"/staffnotes/" + form.targetusername
```

templates/note/message_view.html
```diff
@@ -32,6 +32,17 @@
 
     <div class="formatted-content">$:{MARKDOWN(query['content'])}</div>
 
+    $if myself['is_mod']:
+      <form name="copynotetostaffnotes" action="/modcontrol/copynotetostaffnotes" method="post" class="form clear">
```
charmander: Form doesn’t need a name.

templates/note/message_view.html
```diff
@@ -32,6 +32,17 @@
...
+        <button type="submit" class="button positive" style="float: right;">Copy Note to Staff Notes</button>
```
charmander: `type="submit"` is the default.

weasyl/controllers/moderation.py
```diff
@@ -195,16 +196,20 @@ def modcontrol_edituserconfig_(request):
...
 def modcontrol_copynotetostaffnotes_post_(request):
-    form = request.web_input(targetuserid=None, targetusername=None, notecontent=None,
-                             notedatetime=None, notesubject=None)
+    form = request.web_input(noteid=None)
 
-    staffnotetitle = ''.join(("Received note from ", form.targetusername, ", dated ", form.notedatetime,
-                              ', and the subject was: "', form.notesubject, '".'))
+    notedata = note.select_view(request.userid, int(form.noteid))
+
+    staff_note_title = "Received note from {sender}, dated {date}, and the subject was: \"{subj}\".".format(
```
charmander: The mix of omitted and included articles is a bit odd. How about:

> Received note from {sender}, dated {date}, with subject “{subject}”

(Smart quotes there to avoid escaping and because they look nice; add `# encoding: utf-8` and make it a `unicode` literal.)

charmander:
Hmm. GitHub’s UI for review isn’t fantastic there.

templates/note/message_view.html
```diff
@@ -32,6 +32,13 @@
 
     <div class="formatted-content">$:{MARKDOWN(query['content'])}</div>
 
+    $if myself['is_mod']:
+      <form action="/modcontrol/copynotetostaffnotes" method="post" class="form clear">
+        <button class="button positive" style="float: right;">Copy Note to Staff Notes</button>
```
charmander: Also probably doesn’t need a `positive`. There’s already one bold, green, primary button on the page.

charmander:
Thanks!

Issue #172, opened 2016-09-25
Use HTML5 audio player when possible on submission pages


charmander:
Right now we only offer a non-Flash option for in-gallery playback. It would be nice to add this to submission pages, too.

https://forums.weasyl.com/vbulletin/showthread.php?8795-Audio-submissions-still-require-Flash-but-only-on-submission-pages

PR #176, opened 2016-09-27
Use HTML5's audio tag for MP3 playback

kfkitsune:
For #172.

Drops Flash as the playback method for MP3 files, opting instead to use HTML5's ``<audio>`` tag. This should be supported by virtually all browsers. I don't think anything else uses the media player; can we then drop ``/static/mediaplayer`` from the source tree without any issues?

Places where the component files are referenced:

- ``jwplayer.js`` https://github.com/Weasyl/weasyl/blob/7ec29b361bba9cd72c23b8185fecb28a55650aba/templates/common/page_start.html#L36
- ``player.swf`` https://github.com/Weasyl/weasyl/blob/fb97b39b6f0c6fc33e8c8ff6f4151aee25673e4d/templates/detail/submission.html#L93
- ``swfobject.js`` doesn't seem to be referenced in the source (or GitHub's search isn't pulling it up when I search for ``swfobject``)


templates/detail/submission.html
```diff
@@ -87,17 +87,10 @@ <h1 id="detail-title" class="pad-left pad-right">${query['title']} <i>by</i> <a
...
-          controlbar: "bottom",
-          file: $:{INLINE_JSON(submission['display_url'])},
-          icons: false,
-          width: 500,
-          height: 24
-        });
-        </script>
+        <audio controls>
+          <source type="audio/mpeg" src="${submission['display_url']}">
+          Sorry, but your browser does not support HTML5's &lt;audio&gt; tag.
```
charmander:
```html
<p>Your browser doesn’t support audio playback.</p>
```

Mentioning the tag might be unnecessarily technical. =)

templates/detail/submission.html
```diff
@@ -87,17 +87,10 @@ <h1 id="detail-title" class="pad-left pad-right">${query['title']} <i>by</i> <a
...
-          flashplayer: "${CDNIFY('/static/mediaplayer/player.swf')}",
-          controlbar: "bottom",
-          file: $:{INLINE_JSON(submission['display_url'])},
-          icons: false,
-          width: 500,
-          height: 24
-        });
-        </script>
+        <audio controls>
+          <source type="audio/mpeg" src="${submission['display_url']}">
```
charmander: You can just use `<audio type="audio/mpeg" src="…" controls>` if there’s only one source.

Issue #132, opened 2016-08-11
Edit Site Updates


Hendikins:
(Originally filed by Fiz as Redmine issue 465)

Site updates cannot be edited. Add ability to do so, as it is currently a world of pain to make corrections if required.

kfkitsune:
Tracking: This issue was resolved by #183, which superseded #177.

PR #177, opened 2016-10-01
Add ability to edit site updates

kfkitsune:
For #132.

Adds a new admin-only page which allows editing site updates in the same manner as they were initially submitted (i.e., via a Markdown editor).


charmander:
I’m going to do this with a list because we need a list for something else too.

PR #180, opened 2016-10-28
Marketplace (Commission Search)

taedixon:
This set of changes aims to add a commission search feature to Weasyl. This will allow users to find artists who list their current commission status as **open**, have at least one **commission class** defined with a **price**, and match certain search criteria. The new search interface can be found at `/marketplace`.

Searchers can specify a "type" of commission, which will limit results to those with a commission class that has a name at least partially matching the type provided. In order to encourage usage of standardized classes to make searching more effective for both artists and customers, a list of default commission classes is provided based on the most popular commission class types already existing. This list is also used as an option to populate the name of a commission class when it is being set up. Of course, free-form input for commission classes works on both sides of the equation as well.

This also adds a way to set "preference tags" and "will-not-draw" tags, through `/control/editcommissionprices`. These tags allow commission searchers to more easily find artists who specialize in types of content they wish to commission, or to filter out artists who are not willing to draw what they want. The search field relating to these tags in `/marketplace` is labelled "Content"

Searchers can filter results by setting minimum/maximum price values in their preferred currency. Results will be filtered based on prices set on artist's applicable commission classes after being converted to the searcher's currency. In results, values will be displayed in both the currency specified by the artist and the currency specified in the search (if they differ)

Results are ranked on the following criteria:
1. Number of matching "Preference" tags
2. Time of most recent submission
The intent of this is to reward artists who are active with submitting to the site with higher rankings in commission searches, which also indicates that the commissioner is more likely to receive a response in a timely manner.

Current known issues:
- The UI of the marketplace results page could probably use the touch of someone with a gift for good web design.

Notes:
- this change involves a schema upgrade


templates/etc/marketplace.html
```diff
@@ -0,0 +1,94 @@
...
+        $if previndex != None:
+          <a class="button" href="/marketplace?${QUERY_STRING(dict(form, o=previndex))}" rel="prev">Back</a>
+        $if nextindex != None:
+          <a class="button" style="float:right" href="/marketplace?${QUERY_STRING(dict(form, o=nextindex))}" rel="next">Next</a>
+      </div>
+
+    $else:
+      <p>No results found! Try searching with broader terms.</p>
+  </div>
+</div>
```
kfkitsune: Missing a newline here.

weasyl/controllers/marketplace.py
```diff
@@ -0,0 +1,28 @@
+from pyramid.response import Response
```
kfkitsune: Add above this line: ``from __future__ import absolute_import``

weasyl/commishinfo.py
```diff
@@ -20,39 +52,259 @@ def convert_currency(target):
...
+        } for i in query if "a" not in i.settings] + [{
+            "classid": i.classid,
+            "title": i.title,
+            "amount_min": i.amount_min,
+            "amount_max": i.amount_max,
+            "settings": i.settings,
+            "priceid": i.priceid,
+        } for i in query if "a" in i.settings],
         "content": content if content else "",
+        "tags": [i["t"] for i in tags_extract if 'n' not in i["s"]],
```
charmander: Are these the only places `tags_extract` is used? If so, you should be able to write `.fetchall()` on the query and `[i.title for i in tags if 'n' not in i.settings]` here, which should be easier to read.

weasyl/commishinfo.py
```diff
@@ -20,39 +52,259 @@ def convert_currency(target):
...
+                FROM submission sub
+                JOIN searchmapsubmit map ON map.targetid = sub.submitid
+                JOIN searchtag tag ON map.tagid = tag.tagid
+                WHERE tag.title = ANY(%(tags)s)
+                AND sub.rating <= %(rating)s
+                AND sub.settings !~ '[hf]'
+                GROUP BY sub.userid
+            ) AS example ON example.userid = p.userid
+
+            WHERE p.settings ~ '[os]..?'
```
charmander: I don’t really remember how `profile.settings` works, but is this equivalent to `^[os]`?

weasyl/define.py
```diff
@@ -598,24 +598,15 @@ def get_path():
 
 
 def text_price_amount(target):
-    return "%i.%s%i" % (target / 100, "" if target % 100 > 9 else "0", target % 100)
+    return "%.2f" % round(target / 100, 2)
```
charmander: Doesn’t `%.2f` already round? Nice one, anyway!

charmander: I’m also pretty sure integers are passed into this function at some point, and define.py doesn’t have a `from __future__ import division`, so you might need to make this `target / 100.0`.

templates/control/edit_commissionprices.html
```diff
@@ -164,6 +161,43 @@
...
+        $else:
+          <p class="color-lighter">You have not defined any Artist Tags!</p>
+      </div>
+
+      <h3 class="edit">Edit Tags</h3>
+      $:{RENDER("common/detail_tag_form.html", [query['userid'], "user", query['tags']])}
+
+    </div><!-- /di-tags -->
+    <div class="di-tags">
+      <h3>Will-Not-Draw Tags <a class="modify typeface-default color-c" href="#">Modify</a></h3>
```
charmander: Name might be too specific to things that are drawn?

weasyl/commishinfo.py
```diff
@@ -82,7 +333,7 @@ def create_price(userid, price, currency="", settings=""):
 
     # Settings are at most one currency class, and optionally an 'a' to indicate an add-on price.
     # TODO: replace these character codes with an enum.
-    settings = "%s%s" % ("".join(i for i in currency if i in "epycmu")[:1],
+    settings = "%s%s" % ("".join(i for i in currency if i in "epycmufzn")[:1],
```
charmander: Could you change these `"epycmufzn"` to use a constant built from `CURRENCY_CHARMAP`?

weasyl/test/test_marketplace.py
```diff
@@ -0,0 +1,141 @@
...
+    results = commishinfo.select_commissionable(userid=u1,
+                                                limit=10,
+                                                offset=0,
+                                                q="",
+                                                commishclass="",
+                                                min_price=None,
+                                                max_price=None,
+                                                currency='')
+    rids = [r['username'] for r in results]
+    assert len(rids) == 0
```
charmander: `assert not results`?

assets/scss/site.scss
```diff
@@ -4829,3 +4829,28 @@ a.user-icon:hover > span, a.user-icon:focus > span {
 .thumbnail-stack .caption .title {
     font-size: 13px;
 }
+
+/* ------------------------------------ =marketplace -- */
```
charmander: This can go in `assets/scss/pages/_marketplace.scss` without the comment (import it after the other imports at the top). Mind the spaces after `:`, the semicolon on the `margin-left`, and the newline at the end of the file.

libweasyl/libweasyl/models/tables.py
```diff
@@ -664,6 +664,21 @@ def default_fkey(*args, **kwargs):
 Index('ind_searchmapsubmit_targetid', searchmapsubmit.c.targetid)
 
 
+searchmapartist = Table(
+    'searchmapartist', metadata,
+    Column('tagid', Integer(), primary_key=True, nullable=False),
+    Column('targetid', Integer(), primary_key=True, nullable=False),
+    Column('settings', CharSettingsColumn({
```
charmander: Maybe these should be separate tables instead? `artist_preferred_tags` and `artist_non_preferred_tags`, or something like that.

taedixon: What would be the particular benefits of splitting them out? One of the reasons I went with this approach was that I considered it wouldn't be logically possible for someone to have the same tag in both lists, but beyond that I was mostly just mirroring the other searchmap* tables

templates/control/edit_commissionprices.html
taedixon: Perhaps, but we certainly can't call them blacklist tags again :b
Maybe "opt-out"? "avoidance tags"? "the no-no list"?

libweasyl/libweasyl/models/tables.py
taedixon: It turns out that searchtag.associate assumes there will be a settings column (all other searchmap tables use it to just store whether it's an artist-set tag or not), so in the interest of reducing complexity I put them in separate tables that each have an unused settings column. If you'd prefer, I could drop the settings and set up a special case for these tags instead but I assumed that wouldn't be ideal.

The other option, I guess, would be to refactor ALL the searchmaps but I considered that beyond the scope of this change.

templates/control/edit_commissionprices.html
```diff
@@ -164,6 +161,43 @@
 
       <button type="submit" class="button positive" style="float: right;">Save Description</button>
     </form>
+
+    <div class="di-tags">
+      <h3>Preference Tags <a class="modify typeface-default color-c" href="#">Modify</a></h3>
```
charmander: How about “preferred tags”?

templates/control/edit_commissionprices.html
```diff
@@ -164,6 +161,43 @@
...
+        Use these tags to indicate subject matter you prefer to create, making it easier for
+        commissioners to find you!
+      </p>
+
+      <div class="tags open clear">
+        $if query['tags']:
+          $for tag in query['tags']:
+            <a href="/marketplace?q=${tag}">${tag}</a>
+        $else:
+          <p class="color-lighter">You have not defined any Preference Tags!</p>
```
charmander: An exclamation mark and capitalized name here are probably unnecessary.

templates/etc/marketplace.html
```diff
@@ -0,0 +1,94 @@
...
+              <a href="/search?${artist['searchquery']}">[Tagged: ${artist['examplecount']}]</a>
+          </div>
+          $# commission price range
+          <div class="artist-prices">
+            $if artist['pricemin']:
+              <span>$:{'from ' if artist['pricemax'] else ''}$:{SYMBOL(artist['pricesettings'])} ${PRICE(artist['pricemin'])}</span>
+              $if artist['pricemax'] and artist['pricemax'] != artist['pricemin']:
+                <span>to $:{SYMBOL(artist['pricesettings'])} ${PRICE(artist['pricemax'])}</span>
+              $if artist['localmin'] != artist['pricemin']:
+                ( <span>Approx. $:{SYMBOL(form.currency)} ${PRICE(artist['localmin'])}</span>
```
charmander: There shouldn’t be spaces inside the parentheses, even though web.py makes it awkward.

Maybe:

```html
$code:
  if artist['localmin'] != artist['pricemin']:
    local = 'Approx. %s %s' % (SYMBOL(form.currency), PRICE(artist['localmin']))
    if artist['pricemax'] and artist['pricemax'] != artist['pricemin']:
        local += ' to %s %s' % (SYMBOL(form.currency), PRICE(artist['localmax']))
(${local})
```

templates/etc/marketplace.html
```diff
@@ -0,0 +1,94 @@
...
+                $if artist['pricemax'] and artist['pricemax'] != artist['pricemin']:
+                  <span>to $:{SYMBOL(form.currency)} ${PRICE(artist['localmax'])}</span>
+                )
+            $else:
+              Price Not Declared
```
charmander: Lowercase “not declared” for consistency and also because it looks nicer.

templates/etc/marketplace.html
```diff
@@ -0,0 +1,94 @@
...
+          <div class="formatted-content">
+            $if artist.get('description'):
+              $:{MARKDOWN(artist['description'])}
+            $else:
+              No description provided
+          </div>
+        </div>
+
+      <div style="padding-top: 1.4em;" class="pad-left pad-right">
+        $if previndex != None:
```
charmander: `is not` to compare with `None`, usually

weasyl/commishinfo.py
```diff
@@ -20,39 +53,261 @@ def convert_currency(target):
...
+    :return: The conversion ratio
+    """
+    valuecode = _charmap_to_currency_code(valuecode)
+    targetcode = _charmap_to_currency_code(targetcode)
+    rates = _fetch_rates()
+    if not rates:
+        # in the unlikely event of an error, invalidate our rates
+        # (to try fetching again) and return value unaltered
+        _fetch_rates.invalidate()
+        return 1.0
```
charmander: This doesn’t seem like the right return value for failure.

weasyl/commishinfo.py
```diff
@@ -20,39 +53,261 @@ def convert_currency(target):
...
+    query = d.engine.execute("".join(stmt), limit=limit, min=min_price,
+                             max=max_price, cclass=commishclass, tags=tags,
+                             rating=max_rating, offset=offset)
+
+    def prepare(info):
+        dinfo = dict(info)
+        dinfo['localmin'] = info.convertedmin
+        dinfo['localmax'] = convert_currency(info.pricemax, info.pricesettings, currency)
+        if info.examplecount and tags:
+            dinfo['searchquery'] = "q=user%3A" + info.username + "+%7C" + "+%7C".join(tags)
```
charmander: Might be clearer as:

```python
terms = ["user:" + info.username] + ["|" + tag for tag in tags]
dinfo['searchquery'] = "q=" + urllib.quote(" ".join(terms))
```

weasyl/commishinfo.py
```diff
@@ -10,8 +13,38 @@
...
+    "p": {"code": "GBP", "name": "British Pound Sterling", "symbol": "&#163;"},
+    "y": {"code": "JPY", "name": "Japanese Yen", "symbol": "J&#165;"},
+    "c": {"code": "CAD", "name": "Canadian Dollar", "symbol": "C&#36;"},
+    "m": {"code": "MXN", "name": "Mexican Peso", "symbol": "M&#36;"},
+    "u": {"code": "AUD", "name": "Australian Dollar", "symbol": "A&#36;"},
+    "z": {"code": "NZD", "name": "New Zealand Dollar", "symbol": "NZ&#36;"},
+    "n": {"code": "CNY", "name": "Chinese Yuan", "symbol": "C&#165;"},
+    "f": {"code": "CHF", "name": "Swiss Franc", "symbol": "Fr"},
+}
+ALL_CURRENCIES = "".join([key for key, value in CURRENCY_CHARMAP.items()])
```
charmander: This is `"".join(CURRENCY_CHARMAP.keys())`. Also – sorry for not noticing this in the earlier review – if it’s only used for `in` checks, `CURRENCY_CHARMAP` can be used directly (`"".join(c for c in stuff if c in CURRENCY_CHARMAP)`).

weasyl/commishinfo.py
taedixon: You are probably right, but what *should* be the right value? Raising an exception seems excessive considering the conversion is more of a "nice to have". than "required functionality". None might make sense but it is slightly inconvenient and i'd probably have to assume a value for the sql query anyway. 0 is an obviously erroneous non-null value and I GUESS for truly robust code I should assume it's possible for fixer.io to send 0 or some other malformed value as well.

How it works right now is that it doesn't display the "converted" price if it's equal to the original price, covering both the cases of "they are the same currency" and "disastrous failure of the service we are using". But is that putting too many assumptions into the view? Please advise.

weasyl/commishinfo.py
```diff
@@ -20,39 +53,261 @@ def convert_currency(target):
...
-    content = d.execute("SELECT content FROM commishdesc WHERE userid = %i", [userid], ["element"])
 
+def select_list(userid):
+    query = d.engine.execute("SELECT classid, title, amount_min, amount_max, settings, priceid FROM commishprice"
+                             " WHERE userid = %(userid)s ORDER BY classid, title", userid=userid)
+    classes = d.engine.execute("SELECT classid, title FROM commishclass WHERE userid = %(id)s ORDER BY title", id=userid)
+    content = d.engine.execute("SELECT content FROM commishdesc WHERE userid = %(id)s", id=userid).scalar()
+    preference_tags = d.engine.execute("SELECT DISTINCT tag.title FROM searchtag tag "
+                                       "join artist_preferred_tags pref on pref.tagid = tag.tagid "
+                                       "join login l on l.userid = pref.targetid ", userid=userid).fetchall()
```
taedixon: oh wow I can't believe I didn't notice this in testing. That's definitely an oopsie.

weasyl/controllers/settings.py
```diff
@@ -84,9 +84,12 @@ def control_editprofile_put_(request):
 
 @login_required
 def control_editcommissionprices_(request):
+
```
kfkitsune: This stray blank like looks like it was accidentally added.

libweasyl/libweasyl/alembic/versions/40c00abab5f9_add_artist_tags.py
```diff
@@ -42,6 +42,7 @@ def upgrade():
                     )
     op.create_index('ind_artist_optout_tags_tagid', 'artist_optout_tags', ['tagid'])
     op.create_index('ind_artist_optout_tags_targetid', 'artist_optout_tags', ['targetid'])
+    op.execute('CREATE EXTENSION IF NOT EXISTS FUZZYSTRMATCH')
 
 
 def downgrade():
```
kfkitsune: Companion to the ``upgrade()`` block's additional line for this commit (the original state of the DB should be restored on execution of ``downgrade()``); add the following line at the end of the downgrade function:

``op.execute('DROP EXTENSION IF EXISTS FUZZYSTRMATCH')``

weasyl/commishinfo.py
charmander: I would suggest `None`, yep.

weasyl/commishinfo.py
```diff
@@ -463,6 +463,8 @@ def edit_content(userid, content):
 
 def remove_class(userid, classid):
     d.execute("DELETE FROM commishclass WHERE (classid, userid) = (%i, %i)", [d.get_int(classid), userid])
+    # poor man's cascade
+    d.execute("DELETE FROM commishprice WHERE (classid, userid) = (%i, %i)", [d.get_int(classid), userid])
```
taedixon: Its debatable whether this change belongs in this PR or not to be honest. There was an issue brought up in user acceptance that I couldn't replicate locally so i'd set this to go on testing as a bit of exploratory debugging.  
Essentially, if you remove a class without deleting its prices then create a new class, the old prices will be assigned to it. This can also cause other strange behaviour in rendering the commission prices portion of the user profile. So, I'd consider it at least unintended behaviour.    
While setting up prices is closely related to marketplace I think the scope of handling this change correctly as well as possibly the numerous other issues that could come up while attempting to modernize commission prices should probably justify shifting the commit into #188

charmander:
Some changes suggested in taedixon/Weasyl#1.

templates/etc/marketplace.html
```diff
@@ -0,0 +1,115 @@
...
+              Price not declared
+          </div>
+          $# matching commission classes
+          $if form.c or form.pc:
+            <div class="artist-classes">
+              <strong>Matched: ${artist['class']}</strong>
+            </div>
+          $# commission description
+          <div class="formatted-content marketplace-desc-preview">
+            $if artist.get('description'):
```
charmander: Will the `'description'` key ever not exist?

templates/help/help.html
```diff
@@ -17,6 +17,7 @@
 				<li><a href="/help/ratings">Content Ratings</a></li>
 				<li><a href="/help/google-drive-embed">Google Drive Embedding</a></li>
 				<li><a href="/help/searching">Searching on Weasyl</a></li>
+        <li><a href="/help/marketplace">Using the Marketplace</a></li>
```
charmander: This file appears to use tab indentation; may as well follow that for now.

templates/help/marketplace.html
```diff
@@ -0,0 +1,73 @@
+$:{TITLE("Marketplace", "Help", "/help")}
+
+<div class="content">
+  <div class="constrained">
+
+    <h3>What is the Marketplace?</h3>
+    <div class="formatted-content">
+      <p>The Marketplace is Weasyl’s Commission Search tool. You can use it to find artists who are open for commissions that match your search criteria.</p>
```
charmander: Why is “Commission Search” capitalized?

templates/help/marketplace.html
```diff
@@ -0,0 +1,73 @@
...
+    </div>
+
+    <h3>Searching for Artists</h3>
+    <div class="formatted-content">
+      <p>Marketplace currently supports the following search criteria:</p>
+
+      <ul>
+        <li>Type of Commission (e.g. badge)</li>
```
charmander: “Commission” probably shouldn’t be capitalized here to match “range and currency” below.

templates/etc/marketplace.html
taedixon: If the user sets up their commission prices but does not fill out the "Additional Information" field, the value may be None. If they fill it out then erase it at a later date, it may be an empty string. But in either case, I can't see a situation where the key itself will be missing under the current setup.

etc/SETTINGS.txt
```diff
@@ -100,6 +101,9 @@ y  Japanese yen
 c  Canadian dollar
 m  Mexican peso
 u  Australian dollar
+z  New Zealand dollar
```
hyena: UGH

templates/etc/marketplace.html
```diff
@@ -0,0 +1,115 @@
...
+            <a href="/submissions/${artist['username']}">[Gallery]</a>
+            $if artist['searchquery']:
+              <a href="/search?${artist['searchquery']}">[Tagged: ${artist['examplecount']}]</a>
+          </div>
+          $# commission price range
+          <div class="artist-prices">
+            $if artist['pricemin']:
+              $# display low end of price range in artist's currency
+              <span>$:{'from ' if artist['pricemax'] else ''}${SYMBOL(artist['pricesettings'])} ${PRICE(artist['pricemin'])}</span>
+              $if artist['pricemax'] and artist['pricemax'] != artist['pricemin']:
```
hyena: Thanks for making ranges optional.

templates/help/marketplace.html
```diff
@@ -0,0 +1,73 @@
...
+    </div>
+
+    <h3>Content</h3>
+    <div class="formatted-content">
+      <p>You can search for artists based on the content you wish to commission by entering tags in this field. Artists who have indicated that they prefer creating the content you’re searching for will appear higher in search results, and artists that do not wish to create that content will be excluded from them.</p>
+    </div>
+
+    <h3>Price</h3>
+    <div class="formatted-content">
+      <p>Marketplace searches can have a minimum and/or maximum price, and you can specify your preferred currency.</p>
```
hyena: Nit: and optionally maximum price.

templates/etc/marketplace.html
```diff
@@ -0,0 +1,115 @@
...
+                $# display high end of price range in artist's currency
+                <span>to ${SYMBOL(artist['pricesettings'])} ${PRICE(artist['pricemax'])}</span>
+              $if artist['pricesettings'] != form.currency and artist['localmin']:
+                $if artist['localmax'] and artist['localmax'] != artist['localmin']:
+                  $# display converted prices (range)
+                  (<span>Approx. ${SYMBOL(form.currency)} ${PRICE(artist['localmin'])}</span>
```
hyena: Thanks for showing this separately.

hyena:
Ranges are nice. But what most people are going to want to do is provide add-ons. (e.g. 'additional characters at $5 each'). How do you see people representing that in this implementation?

PR #183, opened 2016-10-29
Site update list and editing

charmander:
Supersedes #177. Includes ORM stuff and a list, with the ability to edit any site update instead of just the most recent one.

Before merging,
- [x] add tests


weasyl/middleware.py
```diff
@@ -141,25 +139,26 @@ def status_check_tween(request):
...
-        request.log_exc()
-        raise HTTPServiceUnavailable("database error")
-
-    # postgresql is still there. Register clean-up of this property.
-    def cleanup(request):
-        db.close()
-
-    request.add_finished_callback(cleanup)
-    return db
+    return d.sessionmaker()
```
hyena: Very nice if this works.

What's the consequence of removing the checking for the database still being there?

The alternative way of doing this is to use a tween that both sets it and removes it. This requires being pretty certain of when db is accessed in the lifetime of the request.

charmander: > What's the consequence of removing the checking for the database still being there?

I’m hoping that it’s “slightly improving response times”. Removing things without knowing why they were added is a little worrying, but it seems reasonable to think that the only thing `SELECT 1` can do is provide a more specific error to the user – something that might be better handled in middleware if it really turns out to be necessary.

libweasyl/libweasyl/models/users.py
```diff
@@ -20,13 +20,21 @@
 log = logging.getLogger(__name__)
 
 
+_BLANK_AVATAR = "/static/images/avatar_default.jpg"
```
hyena: This discussion goes beyond the scope of the code review but:

The existence of libweasyl has made some tooling easier to write (e.g. the thumbnail resize workers). It provides a distinction between the logic that forms a web-application and the logic that manipulates the data.

There is some cross over, but since it's usually in the direction of Weasyl doing something libweasyl should, rather than libweasyl doing something Web Application-like, it's generally remained clean.

PR #184, opened 2016-11-05
Miscellaneous fixes for the Login Email Blacklist

kfkitsune:
- Fixes alignment of text in the columns to be centered to match the centered column headings.
- Eschews the ``d.meta.tables`` method of adding a row, opting for raw SQL to leverage ``ON CONFLICT (domain_name) DO NOTHING`` to prevent errors in the event that a domain name is attempted to be added twice.
- Enforces the 252 character limit on the Python back-end (it was only relying on HTML length limits before).
- ~~Forces domain name to lowercase (because capitalization doesn't really matter)~~ This was already done in the ``d.meta.tables`` method of adding SQL, and was carried over in these changes.

### DIFF
--- a/templates/directorcontrol/emailblacklist.html
+++ b/templates/directorcontrol/emailblacklist.html
@@ -16,10 +16,10 @@ $:{RENDER("common/stage_title.html", ["Blacklist Email Address", "Director Contr
       <tbody>
         $for item in entries:
           <tr>
-            <td><input class="checkbox" type="checkbox" name="remove_selection" value="$item['id']"></td>
-            <td>$item['domain_name']</td>
-            <td>$item['reason']</td>
-            <td>$item['name']</td>
+            <td style="text-align: center;"><input class="checkbox" type="checkbox" name="remove_selection" value="$item['id']"></td>
+            <td style="text-align: center;">$item['domain_name']</td>
+            <td style="text-align: center;">$item['reason']</td>
+            <td style="text-align: center;">$item['name']</td>
           </tr>
       </tbody>
     </table>
@@ -31,7 +31,7 @@ $:{RENDER("common/stage_title.html", ["Blacklist Email Address", "Director Contr
 <div class="content">
   <form method="POST" action="/directorcontrol/emailblacklist" class="form skinny clear">
     <label for="domain_name">Domain Name to Blacklist</label>
-    <input type="text" class="input" maxlength="252" name="domain_name" id="domain_name">
+    <input type="text" class="input" maxlength="252" name="domain_name" id="domain_name" required>
 
     <label for="reason">Reason Why Blacklisted</label>
     <textarea rows="4" cols="60" class="input" name="reason" id="reason"></textarea>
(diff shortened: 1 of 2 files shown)


templates/directorcontrol/emailblacklist.html
```diff
@@ -16,10 +16,10 @@
       <tbody>
         $for item in entries:
           <tr>
-            <td><input class="checkbox" type="checkbox" name="remove_selection" value="$item['id']"></td>
-            <td>$item['domain_name']</td>
-            <td>$item['reason']</td>
-            <td>$item['name']</td>
+            <td style="text-align: center;"><input class="checkbox" type="checkbox" name="remove_selection" value="$item['id']"></td>
```
charmander: How about aligning the headers to the left instead?

weasyl/controllers/director.py
```diff
@@ -36,11 +36,12 @@ def directorcontrol_emailblacklist_post_(request):
...
 
     # Add entry to blacklist
-    elif form.action == "add":
-        d.engine.execute(d.meta.tables["emailblacklist"].insert(), {
-            "domain_name": form.domain_name.lower(),
-            "reason": form.reason,
-            "added_by": request.userid,
-        })
+    elif form.action == "add" and form.domain_name:
+        if len(form.domain_name) <= 252:
```
charmander: Not necessary (especially if you’re not going to produce a useful error); the database field already has a length limit.

charmander:
> Forces domain name to lowercase (because capitalization doesn't really matter).

Didn’t it already do that?

kfkitsune:
@charmander, in short, no, the domain name isn't normalized to lowercase in the codebase for the blacklist checking portion. The account creation process does ``.lower()`` the domain side of the address at ``emailer.py:L25`` as the returned address, but when the email is passed to ``is_email_blacklisted()`` at ``login.py:L162``, while the code does pass in the normalized address, there isn't any further code checking to see if the address is a different case of address in the ``emailblacklist`` table (it only returns a true/false).

So I figured that it would be best to force ``.lower()`` when the address is added; if you'd rather me handle the conversion elsewhere, let me know and I can move it into the ``is_email_blacklisted()`` function, or wherever else you think would be better.

charmander:
This diff has a

```diff
- "domain_name": form.domain_name.lower(),
+ domain_name=form.domain_name.lower(),
```

If there’s another change, I’m not seeing it.

kfkitsune:
@charmander Oh. That. Somewhere in between converting from the ``d.meta.tables`` to raw SQL for the conflict handling, I had forgotten to add back in the ``.lower()``, so in my head I was like, let's force it! I'll edit the PR description. My bad!

templates/directorcontrol/emailblacklist.html
```diff
@@ -1,42 +1,47 @@
...
+              <td>$item['domain_name']</td>
+              <td>$item['reason']</td>
+              <td>$item['name']</td>
+            </tr>
+        </tbody>
+      </table>
+      <button class="button positive" name="action" value="remove">Unblacklist Selected Domains</button>
+    </form>
+  $else:
+    <span class="label" style="text-align: center;">No email domains are currently blacklisted from creating an account.</span>
```
charmander: `<span class="label">` is for things that label inputs but shouldn’t be `<label>`s for some reason. This can be a `<p>`.

PR #185, opened 2016-11-15
Remove unused `files.read` parameters; let errors propagate

charmander:
“Missing submission/journal file” shouldn’t be ignored, and the Markdown renderer shouldn’t be the one to handle it.


hyena:
+1

PR #191, opened 2016-12-31
Fix Vagrantfile database extraction and adjust libweasyl requirements

skylerbunny:


hyena:
We should switch back to gunzip when we have the chance, I believe, but that's something for ikanimew.

PR #192, opened 2017-01-01
Marketplace 

taedixon:
Marketplace - Second Strike


weasyl/searchtag.py
```diff
@@ -167,7 +167,7 @@ def is_tag_restriction_pattern_valid(text):
     return False
 
 
-def associate(userid, tags, submitid=None, charid=None, journalid=None):
+def associate(userid, tags, submitid=None, charid=None, journalid=None, artistid=None, optoutid=None):
```
hyena: This method is becoming monstrous and the way things are being passed into it is imo an anti-pattern. `get_targetid` is also bad.

`artistid` is a bad name because it doesn't suggest what's it for. Consider `preferred_tags_userid` and `optout_tags_userid`.

weasyl/searchtag.py
```diff
@@ -200,14 +204,23 @@ def associate(userid, tags, submitid=None, charid=None, journalid=None):
...
+    elif journalid:
         table, feature = "searchmapjournal", "journal"
         ownerid = d.get_ownerid(journalid=targetid)
+    elif artistid:
+        table, feature = "artist_preferred_tags", "user"
+        targetid = ownerid = artistid
+    elif optoutid:
+        table, feature = "artist_optout_tags", "user"
+        targetid = ownerid = optoutid
+    else:
```
hyena: This is an improvement at least over the old version.

weasyl/searchtag.py
```diff
@@ -227,6 +240,10 @@ def associate(userid, tags, submitid=None, charid=None, journalid=None):
     added = entered_tagids - existing_tagids
     removed = existing_tagids - entered_tagids
 
+    # enforce the limit on artist preference tags
+    if artistid and (len(added) - len(removed) + len(existing)) > 50:
```
hyena: 50 should be moved to a constant near the top of the file (e.g. `MAX_PREFERRED_TAGS = 50` as opposed to hidden in the code.

weasyl/commishinfo.py
```diff
@@ -20,69 +49,294 @@ def convert_currency(target):
...
+        # As well, use searched class as a tag for purposes of finding "tagged" examples of an artists work
+        tags.append(commishclass)
+    stmt.append("""
+            COALESCE(COUNT(preftag.tagid), 0) DESC, latest_submission DESC
+        LIMIT %(limit)s OFFSET %(offset)s
+    """)
+    # to allow partial matches on commishclass
+    cclasslike = "%" + commishclass + "%"
+    max_rating = d.get_rating(userid)
+    print("".join(stmt))
```
hyena: Debug statement left in code.

weasyl/commishinfo.py
```diff
@@ -20,69 +49,294 @@ def convert_currency(target):
...
+    :param max_price: Do not return artists with a minimum price above this value
+    :param currency: The single-character currency code to convert the returned results in,
+                     as defined in commishprice.settings
+    :param offset: Results set offset for pagination. Must be a positive Number
+    :param limit: max number of results to return
+    :return: The resulting list of artists & their commission info
+    """
+    stmt = ["""
+        SELECT p.userid, p.username, p.settings,
+            MIN(cp.amount_min) AS pricemin,
```
hyena: Why is this min necessary?

hyena: Nevermind. I see.

weasyl/searchtag.py
taedixon: I don't disagree but refactoring get_targetid and related is probably an issue all on its own.

point on name changes taken

templates/control/edit_commissionprices.html
```diff
@@ -1,169 +1,268 @@
...
+          <!-- Copied from edit_profile.html. TODO(hyena): Only put this in one place, probably here. -->
+          <label for="editprofile-setcommish">Commissions</label>
+          <select id="editprofile-setcommish" class="input" name="set_commish">
+            <option value="o"${_SELECTED(0, 'o')}>I am currently accepting commissions</option>
+            <option value="s"${_SELECTED(0, 's')}>I may sometimes accept commissions</option>
+            <option value="f"${_SELECTED(0, 'f')}>My commission queue is currently filled</option>
+            <option value="c"${_SELECTED(0, 'c')}>I am not accepting commissions right now</option>
+            <option value="e"${_SELECTED(0, 'e')}>This is not applicable to me</option>
+          </select>
+
```
skylerbunny: It might be nice if at this location in the control panel, there's a comment like 'Setting your status to X will make commission items on this page appear in the [marketplace].

Hendikins:
Since it's now possible to edit all commission-related settings in one place, we should call the settings page "Commission Settings" rather than "(Edit) Commission Prices".

hyena:
- [ ] Fix the name of the preferences pane to be 'Commission Settings'
- [ ] Fix up help images to reflect the new state of things.

PR #193, opened 2017-01-02
Fix searchtag test failures by using more sets

weykent:
>The tests were failing because the order was being compared, when it wasn't actually important.

### DIFF
--- a/weasyl/searchtag.py
+++ b/weasyl/searchtag.py
@@ -439,8 +439,7 @@ def get_user_tag_restrictions(userid):
         WHERE userid = %(userid)s
         ORDER BY st.title
     """, userid=userid).fetchall()
-    tags = [tag.title for tag in query]
-    return tags
+    return {tag.title for tag in query}
 
 
 def get_global_tag_restrictions(userid):
--- a/weasyl/test/searchtag/test_edit_global_tag_restrictions.py
+++ b/weasyl/test/searchtag/test_edit_global_tag_restrictions.py
@@ -9,9 +9,9 @@ from weasyl.error import WeasylError
 from weasyl.test import db_utils
 
 # Test lists of tags
-valid_tags = ['test', '*test', 'te*st', 'test*', 'test_too']
-invalid_tags = ['*', 'a*', '*a', 'a*a*', '*a*a', '*aa*', 'a**a', '}']
-combined_tags = valid_tags + invalid_tags
+valid_tags = {'test', '*test', 'te*st', 'test*', 'test_too'}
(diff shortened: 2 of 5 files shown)


weasyl/searchtag.py
```diff
@@ -439,8 +439,7 @@ def get_user_tag_restrictions(userid):
         WHERE userid = %(userid)s
         ORDER BY st.title
     """, userid=userid).fetchall()
-    tags = [tag.title for tag in query]
-    return tags
+    return {tag.title for tag in query}
```
hyena: This requires an update of the docstring on line 433.

weasyl/test/searchtag/test_edit_global_tag_restrictions.py
```diff
@@ -22,7 +22,7 @@ def test_edit_global_tag_restrictions_fully_clear_entries_after_adding_items(mon
     searchtag.edit_global_tag_restrictions(director_user_id, tags)
     tags = set()
     searchtag.edit_global_tag_restrictions(director_user_id, tags)
-    assert searchtag.get_global_tag_restrictions(director_user_id) == []
+    assert not searchtag.get_global_tag_restrictions(director_user_id)
```
hyena: Nit: I prefer `== {}`  because of the possible weird case where get_global_tag_restrictions returns `None` due to a bug.

weykent: Sadly, `{}` is a dict, not a set. I'll update it though.

weasyl/searchtag.py
weykent: K.

weasyl/test/searchtag/test_edit_global_tag_restrictions.py
hyena: `set([])`?

hyena:
This method looks the same as the one on line 469. What's the intended difference?

Furthermore note that this method returns the tags with `ORDER BY st.title`. So why was the order ever inconsistent to begin with?

hyena:
In some ways the error was actually the test: The docstring for the method specifically states that its purpose is to return a LIST of tags for 'friendly display to the user'. That is, a sorted list in alphabetical order. The parameters passed in are not sorted:
```
Python 2.7.12 (default, Jul  1 2016, 15:12:24) 
[GCC 5.4.0 20160609] on linux2
>>> a = ['test', '*test', 'te*st', 'test*', 'test_too']
>>> a.sort()
>>> a
['*test', 'te*st', 'test', 'test*', 'test_too']
```

It's probably a shed-painting detail as to whether it's better to do a sort of the list in SQL (as it is now), in Python, or in the template (as it is after this change). But if it's not a list, we probably don't need two methods, unless I'm misreading.

PR #196, opened 2017-01-09
Use TIMESTAMP column for latest submission time

charmander:
`WeasylTimestampColumn` is for compatibility with very old code; it shouldn’t be used with new columns.


libweasyl/libweasyl/alembic/versions/eff79a07a88d_use_timestamp_column_for_latest_.py
```diff
@@ -0,0 +1,48 @@
+"""Use TIMESTAMP column for latest submission
+
+Revision ID: eff79a07a88d
+Revises: 7ff3906eccf5
```
hyena: This looks wrong.

charmander: It is! Thank you.

Issue #203, opened 2017-01-17
'Tagged in gallery' uses a user's display name, not username


skylerbunny:
'Tagged in gallery' uses a user's display name, not username. This is an issue when, if you attempt to click on the commission entry for (say) ~midsummeraria, you are taken to the search URL https://www.weasyl.com/search?q=user%3AMidsummer%20Aria (+ search terms). You should instead be taken to https://www.weasyl.com/search?q=user%3Amidsummeraria (+ search terms).

skylerbunny:
Fixed in #204 .

PR #204, opened 2017-01-17
Use LOGIN for building links to users

taedixon:
addresses #203 
convert a user's display name to its canonical format before using it in links.

### DIFF
--- a/templates/etc/marketplace.html
+++ b/templates/etc/marketplace.html
@@ -61,8 +61,8 @@ $code:
           </a>
           $# artist's commission info
           <div class="artist-links">
-            <a href="/~${artist['username']}#user-commissions" class="username">${artist['username']}</a>
-            <a href="/submissions/${artist['username']}">[Gallery]</a>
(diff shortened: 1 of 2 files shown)


templates/etc/marketplace.html
```diff
@@ -61,8 +61,8 @@
           </a>
           $# artist's commission info
           <div class="artist-links">
-            <a href="/~${artist['username']}#user-commissions" class="username">${artist['username']}</a>
-            <a href="/submissions/${artist['username']}">[Gallery]</a>
+            <a href="/~${LOGIN(artist['username'])}#user-commissions" class="username">${artist['username']}</a>
+            <a href="/submissions/${LOGIN(artist['username'])}">[Gallery]</a>
             $if artist['searchquery']:
               <a href="/search?${artist['searchquery']}">[Tagged in Gallery]</a>
```
skylerbunny: This needs to use LOGIN as well. It's what alerted me to this problem in the first place: it produces a search URL like 'user:Monocle Cosplay |story'.

skylerbunny:
+1, now works as designed. Thank you!

Issue #122, opened 2016-08-11
 Mod/ContentByUser improvement: Show content alongside thumbnail 


Hendikins:
(Previously GitLab issue 19)

ContentByUser currently shows the content thumbnail alongside its details. This is fine, as long as the user has not uploaded a custom thumbnail.

If the user has uploaded a custom thumb, there is a risk that the thumbnail may not be representative of the uploaded content. As this is a moderation tool, it should probably show a small version of the actual uploaded content too.

Easiest way to do this is probably to show the generated thumbnail alongside the custom thumbnail.

PR #208, opened 2017-01-17
Display custom thumbnails alongside original in contentbyuser

taedixon:
addresses #122 
Add a column to contentbyuser that shows either the cover image or generated thumbnail of a submission/character as appropriate. Generated thumbnails might cut some things off but they'll reduce the bandwidth of this page a lot and if something is getting cut off it's aspect ration is probably too crazy to see anything anyway.

### DIFF
--- a/templates/modcontrol/contentbyuser.html
+++ b/templates/modcontrol/contentbyuser.html
@@ -35,11 +35,26 @@ $code:
             </div>
           </td>
           <td class="thumb-cell">
-            $if 'sub_media' in i:
-              $ thumb = THUMB(i)
-              <img src="${thumb['display_url']}" alt="" class="thumb" />
+            $# custom thumb, if applicable
+            $if i['contype'] == 10 and 'thumbnail-custom' in i['sub_media']:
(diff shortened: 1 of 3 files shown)


taedixon:
👍 doesn't seem to be any other instances that were missed, looks good to me.

templates/modcontrol/contentbyuser.html
```diff
@@ -35,11 +35,28 @@
...
-              $ thumb = THUMB(i)
-              <img src="${thumb['display_url']}" alt="" class="thumb" />
+            $# custom thumb, if applicable
+            $if i['contype'] == 10 and 'thumbnail-custom' in i['sub_media']:
+              <img src="${i['sub_media']['thumbnail-custom'][0]['display_url']}" alt="" class="thumb" />
+            $elif i['contype'] == 20:
+              $# for characters, all thumbnails show as 'generated'.. even if they are custom..
+              <img src="${i['sub_media']['thumbnail-generated'][0]['display_url']}" alt="" class="thumb" />
+            $elif i['contype'] == 30:
+              &nbsp;
```
charmander: Is this just for the `else`? `$elif i['contype'] != 30:` doesn’t read too badly.

PR #216, opened 2017-01-27
Correct link to commission settings in marketplace

charmander:


taedixon:
👍 doesn't seem to be any other instances that were missed, looks good to me.

PR #222, opened 2017-02-27
Add Two-Factor Authentication (2FA)

kfkitsune:
Adds 2FA as an option when signing into an account. This PR aims to:

- Implement 2FA via a TOTP system, where an authenticator application such as Google Authenticator or FreeOTP can be used to generate time-based OTPs.
- Deliver the TOTP provisioning information via a QRcode for ease of loading into the user's chosen authenticator program. Additionally, provides the 2FA secret in text form for use when the authenticator or device does not support scanning the provided QRcode.
- Permit a user who has misplaced or lost their authenticator to regain access to their account via a recovery code system.
- Prevent lockout of accounts when a user intentionally or inadvertently uses all recovery codes for login purposes by force-disabling 2FA if a total of zero recovery codes remain after successful authentication.
- Allow regeneration of the allotment of recovery codes upon proving they are the account owner via a TOTP or recovery code.
- Allow disabling 2FA after providing a TOTP or recovery code.
- Store the recovery codes securely by bcrypt'ing the recovery code prior to storage in the database.
- Store the TOTP code securely by symmetrically encrypting the TOTP code prior to storing it in the user's ``login`` record (encryption key is stored in ``site.config.txt``; a development-specific encryption key has been pushed to the ``site.config.txt.example`` file, and is suitable for development purposes without further action (AKA, change the key if placing it into production environments)).

**Recovery Code Information**
As coded in this PR, the code length was chosen at 20 characters in length, with a key space defined ``string.ascii_uppercase + "123456789"``; numeral zero (0) is specifically excluded because of potential confusion over uppercase 'O' versus numeral '0' in some fonts. Overall strength of the recovery code can therefore be considered to be approximately ``(35^20)/2``, if we are considering the brute-force resistance; this works out to be about 3.8 nonillion possible combinations until the half-way point of possible permutations.

For comparison, Google runs their recovery code system with 8 numeric digits ``10^8``, or fifty million permutations to the half-way mark.

**Sign in process**
Upon the user's knowledge-factor being authenticated, the user is not fully logged in, but the fact that they authenticated and the time that they authenticated is saved on the user's session. From that point, there is a five-minute period for the possession-factor (token/recovery code) to be authenticated for a successful login, or the user's session will be reset to the logged out state.

**New Python Package Requirements**

``pyotp``, ``qrcodegen``, and ``cryptography`` have been added as requirements to support 2FA. ``pyotp`` handles TOTP creation/verification, and ``qrcodegen`` generates QRcodes. ``cryptography``, meanwhile, encrypts the TOTP key prior to storing it into the user's login record.

**Operating system package requirements**
**Note**: These requirements seem to be satisfied already by the development environment. I am leaving these here for documentation purposes, as I discovered that when I tried installing the ``cryptography`` on my local machine (not the vagrant-env), the requirements were needed; so just as a basic heads up and info.

From https://cryptography.io/en/latest/installation/#building-cryptography-on-linux --
```
cryptography should build very easily on Linux provided you have a C compiler, headers for Python (if you’re not using pypy), and headers for the OpenSSL and libffi libraries available on your system.

For Debian and Ubuntu, the following command will ensure that the required dependencies are installed (replace python-dev with python3-dev if you’re using Python 3):

$ sudo apt-get install build-essential libssl-dev libffi-dev python-dev
```

**New site.config.txt section requirements**
The output of Fernet.generate_key() will need to be saved in the ``[two_factor_auth]`` section of ``site.config.txt``. This needs to be kept secret, as all future encryptions of the TOTP keys will rely on this.

```
kyra@kitty:~/git/weasylDev/weasyl$ python
Python 2.7.9 (default, Jun 29 2016, 13:08:31)
[GCC 4.9.2] on linux2
>>> from cryptography.fernet import Fernet
>>> Fernet.generate_key()
'GlkTNxVzaFyfv6d_Fc2BHZvR-F6QK1RNdm-9P_KzcVs='
>>>
```
So, as an example:
```
<<in site.config.txt>>
[two_factor_auth]
 # This key MUST be changed when in production;
 # See https://cryptography.io/en/latest/fernet/ -- Fernet.generate_key()
 secret_key = GlkTNxVzaFyfv6d_Fc2BHZvR-F6QK1RNdm-9P_KzcVs=
```

**Schema**

This **does** require a schema change. A new table named ``twofa_recovery_codes`` has been added to store recovery codes, and ``login`` has been modified to add a column named ``twofa_secret``, which corresponds to the 2FA secret for a given user.

**Remaining items**

- [x] ~~Figure out proper verbiage for use in the various templates. I don't know how the Directors want to handle the situation where someone loses their authenticator as well as recovery codes. Perhaps extend as an admin or director-level function to disable 2FA for a given account? A force-disable function does exist, so this could be feasible to implement in the future.~~
==> The consensus at this time is to not have a side-channel for admins to disable 2FA in this instance; it is a security function, and the recovery code out is designed to be used if the authenticator is lost.
- [x] ~~Figure out correct visual layout. I am not a CSS or HTML5 layout-guru. The layout as it currently stands is usable.~~
==> The design should be pretty good/usable. But alternate styling can be explored in an alternative PR, if necessary.
- [x] Did I forget any edge cases in my validation for the 2FA functions? I am moderately sure I caught everything a user might try to ... creatively lock themselves out of their account by mucking with the hidden form fields when setting up 2FA.
==> As far as I can tell, all edge cases should be handled.

----
Edit on 20170410: Corrected information about the config file section, the keyspace of the recovery codes, and added information about the hashed recovery codes and encrypted TOTP key.


templates/control/2fa/disable.html
```diff
@@ -0,0 +1,29 @@
...
+$:{RENDER("common/stage_title.html", ["Disable 2FA", "Two-Factor Authentication", "/control/2fa/status"])}
+
+<div class="content form clear">
+  <form method="POST" class="form skinny clear" action="/control/2fa/disable">
+    $if error == "2fa":
+      <div id="signin_error"><strong>Whoops!</strong> The 2FA response token or recovery code you provided did not validate.</div>
+    $elif error == "verify":
+      <div id="signin_error"><strong>Hey!</strong> Did you want to disable 2FA? You didn't check the verification checkbox!</div>
+    <h3>Disable Two-Factor Authentication (2FA)</h3>
+    <p><strong>Username: $:{username}</strong></p>
```
charmander: Why is the username unescaped?

templates/control/2fa/disable.html
```diff
@@ -0,0 +1,29 @@
...
+    <p>
+      You are about to disable 2FA for the above account, and remove the
+      security which it provides. If you wish to re-enable 2FA, you will
+      need to go through through the enabling process once again.
+    </p>
+
+    <label for="disable-tfa">2FA token or Recovery Code</label>
+    <input type="text" class="input" maxlength="20" id="disable-tfa" name="tfaresponse" autocomplete="off" required />
+    <div style="padding-top: 1em;">
+      <label><input class="checkbox" type="checkbox" name="verify" required />I confirm that I want to disable 2FA</label>
```
charmander: There should probably be a space after this `<input>`.

weasyl/test/test_two_factor_auth.py
```diff
@@ -0,0 +1,305 @@
...
+    # This /should/ be zero right now, but verify this for test environment sanity.
+    assert 0 == d.engine.scalar("""
+        SELECT COUNT(*)
+        FROM twofa_recovery_codes
+        WHERE userid = (%(userid)s)
+    """, userid=user_id)
+    assert tfa.get_number_of_recovery_codes(user_id) == 0
+    d.engine.execute("""
+        INSERT INTO twofa_recovery_codes (userid, recovery_code)
+        VALUES ( (%(userid)s), (%(code)s) )
```
charmander: These parentheses seem superfluous. `VALUES (%(userid)s, %(code)s)`

weasyl/two_factor_auth.py
```diff
@@ -0,0 +1,340 @@
...
+        userid: The userid to compare the 2FA challenge-response against.
+        tfa_response: User-supplied response. May be either the Google Authenticator
+        (or other app) supplied code, or a recovery code
+        consume_recovery_code:  If set to True, the recovery code is consumed if it is
+        valid, otherwise the call to this function is treated as a query to the validity
+        of the recovery code. (Default: Boolean True)
+
+    Returns: Boolean True if 2FA verification is successful, Boolean False otherwise.
+    """
+    # Length six (6) is a TOTP code...
```
charmander: Define `LENGTH_TOTP_CODE` and `LENGTH_RECOVERY_CODE`?

weasyl/two_factor_auth.py
```diff
@@ -0,0 +1,340 @@
...
+        userid: The userid to save the recovery codes for.
+        recovery_codes: Comma separated unicode string of recovery codes.
+
+    Returns: Boolean True if the codes were successfully saved to the database, otherwise
+    Boolean False
+    """
+    # Force the incoming string to uppercase, then split into a list
+    codes = recovery_codes.upper().split(',')
+    # The list must exist and be equal to the current codes to generate
+    if not len(codes) == _TFA_RECOVERY_CODES:
```
charmander: `!=`

weasyl/two_factor_auth.py
```diff
@@ -0,0 +1,340 @@
...
+        userid: The userid to check for 2FA being enabled.
+
+    Returns: Boolean True if 2FA is enabled for ``userid``, otherwise Boolean False.
+    """
+    result = d.engine.scalar("""
+        SELECT twofa_secret
+        FROM login
+        WHERE userid = (%(userid)s)
+    """, userid=userid)
+    if result:
```
charmander: `SELECT twofa_secret IS NOT NULL`; `return d.engine.scalar(…`

weasyl/two_factor_auth.py
```diff
@@ -0,0 +1,340 @@
...
+        userid: The user for which 2FA should be disabled.
+        tfa_response: User-supplied response. May be either the Google Authenticator
+        (or other app) supplied code, or a recovery code.
+
+    Returns: Boolean True if 2FA was successfully disabled, otherwise Boolean False if the
+    verification of `tfa_response` failed (bad challenge-response or invalid recovery code).
+    """
+    # Sanity checking for length requirement performed in verify() function (6 or 20 length)
+    if verify(userid, tfa_response):
+        d.engine.execute("""
```
charmander: Reuse `force_deactivate` here?

weasyl/controllers/two_factor_auth.py
```diff
@@ -0,0 +1,273 @@
...
+@login_required
+@token_checked
+def tfa_init_qrcode_post_(request):
+    # Return an error if 2FA is already enabled (there's nothing to do in this route)
+    _error_if_2fa_enabled(request.userid)
+
+    # Otherwise, process the form
+    if request.params['action'] == "continue":
+        # Strip any spaces from the TOTP code (some authenticators display the digits like '123 456')
+        tfaresponse = re.sub("[\ ]", '', request.params['tfaresponse'])
```
charmander: `request.params['tfaresponse'].replace(' ', '')`?

libweasyl/libweasyl/models/tables.py
```diff
@@ -391,11 +391,22 @@ def default_fkey(*args, **kwargs):
...
 
 
+twofa_recovery_codes = Table(
+    'twofa_recovery_codes', metadata,
+    Column('userid', Integer(), primary_key=True, nullable=False),
+    Column('recovery_code', String(length=20), primary_key=True, nullable=False),
+    default_fkey(['userid'], ['login.userid'], name='twofa_recovery_codes_userid_fkey'),
+)
+
+Index('ind_twofa_recovery_codes_userid')
```
kfkitsune: ``Index('ind_twofa_recovery_codes_userid', twofa_recovery_codes.c.userid)``

Need to specify the column for the index.

templates/control/2fa/generate_recovery_codes.html
```diff
@@ -0,0 +1,51 @@
+$def with (tfa_recovery_codes, error)
+$:{RENDER("common/stage_title.html", ["Generate Recovery Codes", "Two-Factor Authentication", "/control/2fa/status"])}
+
+<div class="content form clear">
+  <form method="POST" class="form skinny clear" action="/control/2fa/generate_recovery_codes">
+    $if error == "2fa":
+      <div id="signin_error">
+        <strong>Whoops!</strong> The 2FA response token or recovery code you provided did not validate.<strong>
+        If entered, your recovery code was <strong>not</strong> consumed.
```
kfkitsune: ``If entered, your recovery code was <strong>not</strong> consumed.``

Redundant line. If the code didn't validate, it wasn't correct, and therefore it wouldn't be consumed to begin with.

templates/control/2fa/disable.html
```diff
@@ -0,0 +1,29 @@
...
+$:{RENDER("common/stage_title.html", ["Disable 2FA", "Two-Factor Authentication", "/control/2fa/status"])}
+
+<div class="content form clear">
+  <form method="POST" class="form skinny clear" action="/control/2fa/disable">
+    $if error == "2fa":
+      <div id="signin_error"><strong>Whoops!</strong> The 2FA response token or recovery code you provided did not validate.</div>
+    $elif error == "verify":
+      <div id="signin_error"><strong>Hey!</strong> Did you want to disable 2FA? You didn't check the verification checkbox!</div>
+    <h3>Disable Two-Factor Authentication (2FA)</h3>
+    <p><strong>Username: $:{MARKDOWN_EXCERPT(username)}</strong></p>
```
charmander: Why is this using `MARKDOWN_EXCERPT`?

templates/control/2fa/init_verify.html
```diff
@@ -0,0 +1,56 @@
...
+    <p>
+      <strong>Print these codes, and secure them as you would your password.</strong>
+      In the event you lose access to your authenticator app, you will need these
+      codes to regain access to your Weasyl account. Recovery codes can be
+      refreshed at any time from your settings page.
+    </p>
+    <p>
+      <strong>Warning:</strong> For security reasons, Weasyl staff will be unable to assist you if
+      you lose access to your recovery codes. We advise never using more than eight codes, as you will
+      need two to disable 2FA from a logged-out state (one code to log in, and one code to disable
```
charmander: Sounds like you only need one as things are: log in with it, and 2FA will be disabled.

templates/control/2fa/status.html
```diff
@@ -0,0 +1,39 @@
...
+        You currently have ${tfa_recovery_codes_count} valid recovery code${'s' if tfa_recovery_codes_count > 1 else ''} assigned to your account.
+        We advise not relying on recovery codes as a routine access method, and further advise to never
+        use the last two codes; a minimum of two codes are required to disable 2FA from a logged-out state.
+      </p>
+      <p>
+        <a href="/control/2fa/generate_recovery_codes">Click here to get new recovery codes</a>
+      </p>
+    $else:
+      <p>
+        2FA is currently <b>disabled</b>.<br /><br />
```
charmander: Why does this status have two `<br />`s when “enabled” doesn’t?

templates/control/2fa/status.html
```diff
@@ -0,0 +1,39 @@
...
+      factors--you or an adversary who has compromised your password will not be able to fully
+      authenticate to your account.
+    </p>
+    <h3>2FA Status</h3>
+    $if tfa_enabled:
+      <p>
+        2FA is currently <b>enabled</b>.
+      </p>
+      <p>
+        <a href="/control/2fa/disable">Click here to disable 2FA</a>
```
charmander:
```html
<a class="button negative">Disable 2FA</a>
```

templates/control/2fa/init_verify.html
kfkitsune: Technically for this one I didn't want users to rely on using all codes to disable 2FA, as it isn't the "intended" way; I did this more as a mechanism to prevent user's from locking themselves out if all codes were used. That said, I'll just remove the sentence in its entirety, since while the proper path has a user disable from a logged out state with two (login, then deactivate), there is also the backup path of logging in with one code remaining force-disables (and there's no other way to use a code while logged in except to regenerate codes (problem solved), or disable 2FA (also problem solved)).

templates/control/2fa/disable.html
kfkitsune: You asked to have the username escaped through the Markdown renderer. Straight ``$:{MARKDOWN(username)}`` was mucking with the formatting of the text (I'll go back and get a screencap to make sure it still presents the issue). So I tested the ``MARKDOWN_EXCERPT`` mechanism, and it wasn't causing issues with the text, so I used it.

I'll grab a screenshot shortly.

kfkitsune: This is why I am using ``$:{MARKDOWN_EXCERPT(username)}`` versus using ``$:{MARKDOWN(username)}``. The excerpt version isn't giving me the additional carriage return in this instance, and frankly I am unsure why it is doing that.

screenshot from 2017-03-12 19 51 58

charmander: No, I asked to have it escaped. `${username}` rather than `$:{username}`.

kfkitsune: Oh! Got it. I was unaware that was the difference between ``${foo}`` and ``$:{foo}``. Noted for the future. Apologies if there seemed to be any confusion there.

weasyl/controllers/user.py
```diff
@@ -42,6 +43,21 @@ def signin_post_(request):
         # Invalidate cached versions of the frontpage to respect the possibly changed SFW settings.
         index.template_fields.invalidate(logid)
         raise HTTPSeeOther(location=form.referer)
+    elif logid and logerror == "2fa":
+        # Password authentication passed, but user has 2FA set, so verify second factor (Also set SFW mode now)
+        if form.sfwmode == "sfw":
+            request.set_cookie_on_response("sfwmode", "sfw", 31536000)
+        index.template_fields.invalidate(logid)
+        # Store the authenticated userid & password auth time to the session
```
kfkitsune: This may be me being insanely careful about this, but maybe check if a zero-recovery code situation exists here? In theory it should *never* happen (we catch the zero situation down in ``signin_2fa_auth_post_(request)``); but this is me considering a situation where some database lock will mess someone up, even though yes, I know PSQL actions are atomic. Call me paranoid.

Anyway, pseudocode:
```
[[[ Here, in weasyl/controllers/user.py ]]]
remaining_recovery_codes = two_factor_auth.get_number_of_recovery_codes(logid)
if remaining_recovery_codes == 0:
    # The user's account (inexplicably) has zero recovery codes remaining. Force-disable 2FA.
    two_factor_auth.force_disable(logid)
    raise WeasylError("TwoFactorAuthentication-ZeroRecoveryCodesRemaining")  # Note to self, why do error codes not seem to use dashes?

[[[ Over in weasyl/errorcode.py ]]]
"TwoFactorAuthentication-ZeroRecoveryCodesRemaining": "Your account has zero recovery codes remaining, and has had 2FA disabled to prevent you from being permanently unable to log into your account. You may re-enable 2FA if you desire to do so."

[[[ Down in signin_2fa_auth_post_(request) where the zero code check happens ]]]
Do the same thing. Raise the error versus using the custom message?
```

Just my thoughts for now.

charmander: When something that should be impossible happens, defaulting to disabling two-factor authentication and raising an error that isn’t reported via Sentry doesn’t seem like a good idea. Use a `ValueError` or something instead.

weasyl/two_factor_auth.py
```diff
@@ -0,0 +1,344 @@
...
+    # Force the incoming string to uppercase, then split into a list
+    codes = recovery_codes.upper().split(',')
+    # The list must exist and be equal to the current codes to generate
+    if len(codes) != _TFA_RECOVERY_CODES:
+        return False
+    # Make sure all codes are `LENGTH_RECOVERY_CODE` characters long, as expected
+    for code in codes:
+        if len(code) != LENGTH_RECOVERY_CODE:
+            return False
+
```
kfkitsune: I was thinking about these recovery codes being stored plaintext, and this answered the question I had here... Draft NIST SP 800-63B, "5.1.2.2. Look-up Secret Verifiers":

_"Verifiers SHALL store look-up secrets_[Note 1] _in a form that is resistant to offline attacks. Secrets SHALL be hashed with a salt value using an approved hash function as described in [SP 800-132]. The salt value SHALL be a 128-bit or longer random value generated by an approved random bit generator that is stored along with the hash result. A keyed hash function (e.g., HMAC [FIPS198-1]), with the key stored separately from the hashed authenticators (e.g., in a hardware security module) SHOULD be used to further resist dictionary attacks against the stored hashed authenticators."_
**Note 1)** _A common application of look-up secrets is the use of "recovery keys" stored by the subscriber for use in the event another authenticator is lost or malfunctions._

TL,DR: Unless you think otherwise, @charmander, note to self: These probably should be run through a bcrypt cycle. At minimum, this would require the following:

1. In ``store_recovery_codes()``: After pre-checking to ensure the codes are valid, run the codes through bcrypt to hash them before storing them. Optionally, reduce the work-factor slightly, but I don't know how much of a trade-off we would want or need here, keeping in mind the changes needed in ``is_recovery_code_valid()``.
2. In ``is_recovery_code_valid()``: Instead of straight up deleting the code if it is valid and returning it, we'd need to retrieve all current codes, then attempt to verify the provided recovery code in turn against all bcrypt hashes retrieved from the table. Upon finding a match, delete that bcrypt value/userid pair. [This might be a good case for slightly reducing the work-factor, since I am not quite certain how long it takes to run through ten bcrypt cycles back-to-back (worst case).]
3. In the ``twofa_recovery_codes`` table (both the model and the Alembic revision): The length of ``recovery_code`` would need to be extended to match the length of the bcrypt hash (I'd likely just steal the length from ``authbcrypt.hashsum``, which is 100).

The following is recommended as well:
1. In ``weasyl/controllers/two_factor_auth.py:tfa_init_verify_post_()``: This function may need to be looked at to check about the timing of storing the recovery codes versus the ``tfa.activate()`` call. Alternatively, in this module in ``activate()``, change the ``if totp.verify(tfa_response)`` call (L117 as of this comment), to ``if totp.verify(tfa_response, valid_window=1)``, or at most 2 to give a buffer window for the bcrypt iterations over the recovery codes to complete.

I am going to ignore the bit where it says only PBKDF should be used as the hash derivation function (SP 800-132), as we are not bound to use NIST guidelines; I would, however, say that salt-hashing these is likely a Very Good Idea™.

As a side benefit, if we opt to change the recovery code length in the future, no further logic would be required to handle the updated length. Rather, we'd just need to alter the length of newly generated codes, and since input is hashed, we don't really care about input length.

Reference:
https://pages.nist.gov/800-63-3/sp800-63b.html#sec5

weasyl/two_factor_auth.py
```diff
@@ -0,0 +1,349 @@
...
+from qrcodegen import QrCode
+
+from libweasyl import security
+from weasyl import define as d
+from weasyl import login
+
+# Number of recovery codes to provide the user
+_TFA_RECOVERY_CODES = 10
+# Work factor for bcrypt to be used in hashing each recovery code
+BCRYPT_WORK_FACTOR = 10
```
kfkitsune: As a note, I am not sure how much time we want to spend during the setup to generate the bcrypt hashes; I will be happy to change it to a higher value if anyone thinks it should be done (bearing in mind that the main TOTP secret is still unencrypted in ``login``). The overall initial cost would be however long it takes to generate the hash at N-work factor, times the recovery codes provided, which would make a bit of a sizable page load delay during the ``tfa_init_verify_post_()`` controller step, at least depending on how much we want to secure these.

As a comparison, the default work factor bcrypt uses is 12, and Weasyl passwords are 14 (as of this commit, and controlled by a variable in the ``macro.py`` module).

I'll leave it up to the reviewers and other Weasyl staff to determine how long of a page-load is tolerable (though in this case it may be justified, especially if we make note of it in the page itself).

weasyl/two_factor_auth.py
```diff
@@ -0,0 +1,383 @@
...
+    codes = recovery_codes.upper().split(',')
+    # The list must exist and be equal to the current codes to generate
+    if len(codes) != _TFA_RECOVERY_CODES:
+        return False
+    # Make sure all codes are `LENGTH_RECOVERY_CODE` characters long, as expected
+    for code in codes:
+        if len(code) != LENGTH_RECOVERY_CODE:
+            return False
+
+    # Store the recovery codes securely by hashing them with bcrypt (as login.passhash())
```
kfkitsune: For when the inevitable code-review comments come down: s/ (as login.passhash())// -- Just noticed this fragment remained from when I opted to use the slightly reduced work-factor.

weasyl/two_factor_auth.py
```diff
@@ -0,0 +1,383 @@
...
+
+    Character set is defined by `libweasyl.security`, (ASCII+Numbers), limited to uppercase via
+    .upper() for readability.
+
+    Parameters: None
+
+    Returns: A set of length `_TFA_RECOVERY_CODES` where each code is `LENGTH_RECOVERY_CODE`
+    characters in length.
+    """
+    return {security.generate_key(LENGTH_RECOVERY_CODE).upper() for i in range(_TFA_RECOVERY_CODES)}
```
charmander: This biases the key towards letters, reducing its strength from 119 bits to 102 bits. I’d fix it by changing `security.generate_key` to accept an optional `key_characters` defaulting to `string.ascii_letters + string.digits`.

weasyl/two_factor_auth.py
```diff
@@ -0,0 +1,383 @@
...
+
+        INSERT INTO twofa_recovery_codes (userid, recovery_code)
+        SELECT %(userid)s, unnest(%(tfa_recovery_codes)s);
+
+        COMMIT;
+    """, userid=userid, tfa_recovery_codes=list(hashed_codes))
+
+    # Verify if the atomic transaction completed; if `code` (one of the new recovery codes) is
+    #   valid at this point, the new codes were added
+    if is_recovery_code_valid(userid, code, consume_recovery_code=False):
```
charmander: How would they not be added?

config/site.config.txt.example
```diff
@@ -25,3 +25,8 @@ url = postgresql+psycopg2cffi:///weasyl
 [recaptcha-lo.weasyl.com]
 public_key = 6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI
 private_key = 6LeIxAcTAAAAAGG-vFI1TnRWxMZNFuojJ4WifJWe
+
+[twofactorauth.totpsecret.encryption.key]
```
charmander: `[two_factor]` would be a simpler name.

libweasyl/libweasyl/models/tables.py
```diff
@@ -391,11 +391,22 @@ def default_fkey(*args, **kwargs):
...
+    Column('twofa_secret', String(length=420), nullable=True),
 )
 
 Index('ind_login_login_name', login.c.login_name)
 
 
+twofa_recovery_codes = Table(
+    'twofa_recovery_codes', metadata,
+    Column('userid', Integer(), primary_key=True, nullable=False),
+    Column('recovery_code', String(length=100), primary_key=True, nullable=False),
```
charmander: Since this is now hashed, it shouldn’t be a primary key. You may also want to name it to match (`recovery_code_hash`).

weasyl/test/test_two_factor_auth.py
```diff
@@ -0,0 +1,298 @@
...
+        WHERE userid = %(userid)s
+    """, userid=user_id, tfas=tfa_secret_encrypted)
+
+
+@pytest.mark.usefixtures('db')
+def test_get_number_of_recovery_codes():
+    user_id = db_utils.create_user()
+
+    # This /should/ be zero right now, but verify this for test environment sanity.
+    assert 0 == d.engine.scalar("""
```
charmander: Isn’t this the same as the assertion after it?

kfkitsune: Technically, yes; in this case I am just validating my assertion that ``tfa.get_number_of_recovery_codes(user_id)`` returns zero as expected if there are no recovery codes for ``user_id``. In this case it is more of a "Verify that the state of the database is correct in order to validate the following assertion that the function returns properly when the database has zero rows.".

I can remove it if you want, but this was more of an assertion to test my test environment first, before verifying that it returns correctly when actually called.

weasyl/controllers/two_factor_auth.py
```diff
@@ -0,0 +1,251 @@
...
+    elif not verify_checkbox:
+        return Response(define.webpage(request.userid, "control/2fa/disable.html",
+                        [define.get_display_name(request.userid), "verify"]))
+
+
+@login_required
+@twofactorauth_enabled_required
+def tfa_generate_recovery_codes_get_(request):
+    recovery_codes = tfa.generate_recovery_codes()
+    _set_recovery_codes_on_session(','.join(recovery_codes))
```
hyena: Run me through how this situation works with session management?

1. A user opens a browser window and navigates to the screen to generate new recovery codes. They see a list of codes, call it A.
2. They open a second browser window and navigate to the screen to generate new recovery codes there as well. They see a different set of codes. Call it B.
3. They go back to the first browser window and verify that they want to save the recovery codes. They see A on the screen. Which set of codes get saved?

My reading of it is that at step 3 we save a new set of codes inside the user's session. There's one session per cookie so if these actions are done on the same browser, the set of codes that will be saved are B even though it's verified on a screen where A is shown.

On the other hand, if the user uses separate browsers for this (e.g. FireFox and Chromium) with different cookies, then A will be saved.

Although I don't think this will come up too much in practice, it is a little worrisome. The way I could see it being an issue is if a user manages to reload the page at the same time they're issuing their verification post.

weasyl/controllers/user.py
```diff
@@ -42,6 +46,32 @@ def signin_post_(request):
...
+        if form.sfwmode == "sfw":
+            request.set_cookie_on_response("sfwmode", "sfw", 31536000)
+        index.template_fields.invalidate(logid)
+        # Check if out of recovery codes; this should *never* execute normally, save for crafted
+        #   webtests. However, check for it and log an error to Sentry if it happens.
+        remaining_recovery_codes = two_factor_auth.get_number_of_recovery_codes(logid)
+        if remaining_recovery_codes == 0:
+            raise RuntimeError("Two-factor Authentication: Count of recovery codes for userid " +
+                               str(logid) + " was zero upon password authentication succeeding, " +
+                               "which should be impossible.")
```
hyena: Can we add, `Please contact support@weasyl.com`?

weasyl/controllers/user.py
```diff
@@ -65,6 +95,101 @@ def signin_post_(request):
...
+        # Hinder brute-forcing the 2FA token or recovery code by enforcing an upper-bound on 2FA auth attempts.
+        _cleanup_2fa_session()
+        return Response(define.errorpage(
+            request.userid,
+            errorcode.error_messages['TwoFactorAuthenticationAuthenticationAttemptsExceeded'],
+            [["Sign In", "/signin"], ["Return to the Home Page", "/"]]
+        ))
+    else:
+        # Log the failed authentication attempt to the session and save
+        sess.additional_data['2fa_pwd_auth_attempts'] += 1
```
hyena: @charmander How much do you know about our threading model with Pyramid and tornado? I'm vague on it.

This may be crazy, but what's to stop a user from issuing a hundred requests at once? This looks like a read-modify-write situation. There's not an easy way to mutex it because sessions are created on the way in and written back to the database before the response is returned to the user.

weasyl/two_factor_auth.py
hyena: This will block the time to save recovery codes, correct? i.e. the load time of the 'post' request to save codes and enable 2FA?

See comment about which recovery codes get saved. I just want to be careful here to make sure it's infeasible that a user saves a different set of codes than the ones they printed out.

weasyl/two_factor_auth.py
```diff
@@ -0,0 +1,383 @@
...
+    for code in codes:
+        if len(code) != LENGTH_RECOVERY_CODE:
+            return False
+
+    # Store the recovery codes securely by hashing them with bcrypt
+    hashed_codes = [bcrypt.hashpw(code.encode('utf-8'), bcrypt.gensalt(rounds=BCRYPT_WORK_FACTOR)) for code in codes]
+
+    # If above checks have passed, clear current recovery codes for `userid` and store new ones
+    d.engine.execute("""
+        BEGIN;
```
hyena: Good use of transactions. Thanks.

weasyl/two_factor_auth.py
```diff
@@ -0,0 +1,383 @@
...
+        return False
+    # First extract the bcrypt hashes.
+    recovery_code_hash_query = d.engine.execute("""
+        SELECT recovery_code_hash
+        FROM twofa_recovery_codes
+        WHERE userid = %(userid)s
+    """, userid=userid).fetchall()
+    # Then attempt to hash the input code versus the stored code(s).
+    for row in recovery_code_hash_query:
+        if bcrypt.checkpw(tfa_code.upper().encode('utf-8'), row['recovery_code_hash'].encode('utf-8')):
```
hyena: Just a sanity check: How much processing does this take? How feasible is it for a user to DoS us by issuing requests to recover logins with a bogus code like 11111 11111 11111 11111 repeatedly? Obviously this would only work for accounts for which they had a password.

charmander: Yep, this should definitely be rate-limited, and not just for the DoS potential.

weasyl/controllers/user.py
charmander: The default “unexpected error” page should display something to that effect anyway.

kfkitsune: Charmander is correct, here. Using [this](https://testing-old.weasyl.com/search?within=all&cat=1000&q=test&nextid=test&find=submit) as an example:

```
An unexpected error occurred and your request could not be fully processed.
If you wish to contact support (on the forums or by e-mail) about this error, please include the following information: 
request-id: 180b7a29227440a3933105fe2d634986-2lCZ1kxv
```

Where **on the forums** and **by e-mail** are hyperlinked to the location you would expect them to go to (site/forum support and support@...).

PR #223, opened 2017-02-28
Invalidate all other sessions when a user's password is changed.

kfkitsune:
Addresses #221.

Obtains the current session via ``d.get_weasyl_session()``, then deletes all records from the ``session`` table where records match the ``userid`` for whom we just updated their password, and where the ``sessionid`` in the candidate record does not equal the ``sessionid`` in the session object.

### DIFF
--- a/weasyl/profile.py
+++ b/weasyl/profile.py
@@ -565,7 +565,7 @@ def edit_email_password(userid, username, password, newemail, newemailcheck,
         sess = d.get_weasyl_session()
         d.engine.execute("""
             DELETE FROM sessions
-            WHERE userid = (%(userid)s)
+            WHERE userid = %(userid)s
               AND sessionid != (%(currentsession)s)
         """, userid=userid, currentsession=sess.sessionid)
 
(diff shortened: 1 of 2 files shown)


weasyl/profile.py
```diff
@@ -561,6 +561,14 @@ def edit_email_password(userid, username, password, newemail, newemailcheck,
     if newpassword:
         d.execute("UPDATE authbcrypt SET hashsum = '%s' WHERE userid = %i", [login.passhash(newpassword), userid])
 
+        # Invalidate all sessions for `userid` except for the current one
+        sess = d.get_weasyl_session()
+        d.engine.execute("""
+            DELETE FROM sessions
+            WHERE userid = (%(userid)s)
```
charmander: No parentheses required.

weasyl/profile.py
```diff
@@ -561,6 +561,14 @@ def edit_email_password(userid, username, password, newemail, newemailcheck,
     if newpassword:
         d.execute("UPDATE authbcrypt SET hashsum = '%s' WHERE userid = %i", [login.passhash(newpassword), userid])
 
+        # Invalidate all sessions for `userid` except for the current one
+        sess = d.get_weasyl_session()
+        d.engine.execute("""
+            DELETE FROM sessions
+            WHERE userid = %(userid)s
+              AND sessionid != (%(currentsession)s)
```
charmander: here too

hyena:
lgtm. What's up with the two failing tests, though?

PR #224, opened 2017-02-28
Use float division for Recently Popular calculation

charmander:
Integer division was not intended and breaks tests occasionally.


hyena:
Rubber stamped.

PR #228, opened 2017-03-30
Input Hardening & Validation of Input

kfkitsune:
* Add function to validate claimed integers

define.get_int() is broken, and invalid input is getting through to the point where Sentry is picking up on it. Implements define.get_int() as a sane/safe function, and relies solely on Python primitives to determine type (AKA, return an Integer if it gets through int(float(number)), else return zero.

This, however, cuts down on many shenanigans or triggering of Sentry alerts, or will once all _DEPRECIATED calls are changed over to the current get_int()

Redefines the old/insecure version as define.get_int_DEPRECIATED() at the request of Charmander until we can validate that things won't break.

Sentry issues this should resolve:
https://sentry.weasyldev.com/weasyl/weasyl/issues/1483/ - weasyl/profile.py & HTML maxlength added
https://sentry.weasyldev.com/weasyl/weasyl/issues/1538/ - weasyl/profile.py & HTML maxlength added
https://sentry.weasyldev.com/weasyl/weasyl/issues/1513/ - weasyl/journal.py & HTML maxlength added
https://sentry.weasyldev.com/weasyl/weasyl/issues/1377/ - controllers/general.py
https://sentry.weasyldev.com/weasyl/weasyl/issues/1562/ - controllers/marketplace.py

### DIFF
--- a/templates/control/edit_profile.html
+++ b/templates/control/edit_profile.html
@@ -28,7 +28,7 @@ $code:
       <input type="text" id="editprofile-gender" class="input" name="gender" value="${userinfo['gender']}" maxlength="100" />
 
       <label for="editprofile-country">Location</label>
-      <input type="text" id="editprofile-country" class="input" name="country" value="${userinfo['country']}" maxlength="100" />
+      <input type="text" id="editprofile-country" class="input" name="country" value="${userinfo['country']}" maxlength="50" />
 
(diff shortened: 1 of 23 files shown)


charmander:
The word is “deprecated”, but please don’t change `get_int` anyway.

charmander:
and only `maxlength` attributes are necessary; no point in slicing.

PR #237, opened 2017-05-06
Database cleanup

charmander:
Partially addresses #68; required to make the export script (https://github.com/Weasyl/smallcopy) more reliable.


libweasyl/libweasyl/models/tables.py
```diff
@@ -219,7 +185,7 @@ def default_fkey(*args, **kwargs):
 favorite = Table(
     'favorite', metadata,
     Column('userid', Integer(), primary_key=True, nullable=False),
```
hyena: Do we not need an `autoincrement=False` here too? I assumed that the issue was how SQLAlchemy/Alembic was treating Integer() columns set as a primary_key.

charmander: If it’s a foreign key, SQLAlchemy won’t make it auto-increment. Explicit does tend to be better than implicit, though – will include in a future cleanup with the rest of them.

PR #244, opened 2017-05-16
Collaborative filtering

hyena:
This isn't finished, but there's enough implemented to warrant feedback and testing.

*What's working*
  - Ratings are predicted. You can log in and navigate to `/recommendations` to see recommendations for the currently logged in user.

*What's not done yet*
  - The ability to rate an item is not yet implemented other than favoriting.
  - The recommendation page is a placeholder.
  - Algorithm could use tweaks:
    * Cap the number of items returned per user.
    * Consider biasing towards more recent items.
  - I haven't deleted the .ipynb yet. Just keeping it around for my reference.
  - Do some unit tests for basic recommendation cases.

*Other issues to be aware of*
  - This requires running a migration. Expect this to take a while as it effectively copies the favorites table.
  - I did some timing tests and the biggest consumer of time on my laptop (about 30 seconds) is time taken to read in the pandas dataframe from sql. I think we can cut this to about a fifth of that (about 6 seconds) if we sql `COPY` the table to a .csv file and read it in that way.
  - Until we fix that, you might get a 504 timeout on the first load. Subsequent loads should be rapid.
  - These changes require several additional hundred megabytes of storage due to the bloat of python's scientific libraries. Resident memory hits about 1.5gigs when everything is up and running and the recommendations are calculated.
  - If this is put on testing, it would be best to use the larger database for it.


weasyl/recommendation.py
```diff
@@ -0,0 +1,243 @@
...
+@d.record_timing
+def recs_for_user(userid, k=20, count=100):
+    """
+    Generates recommendations for a weasyl user.
+    Will not include the user's own submissions or items they've rated.
+
+    Args:
+        userid (int): The weasyl userid.
+        k (int, optional): How many closest other users to use. Defaults to 20.
+        count (int, optional): How many items to return for the user. Defaults to 10.
```
kfkitsune: ``Defaults to 10.`` => ``Defaults to 100.`` -- The function definition currently has ``count`` default to 100.

weasyl/test/test_recommendation.py
```diff
@@ -0,0 +1,42 @@
...
+        rec.clear_user_rating(self.userid, self.submitid)
+        assert 0 == self.ratingCount()
+        rec.set_user_rating(self.userid, self.submitid, rec.RecommendationRating.LIKE)
+        rec.set_user_rating(self.userid, self.submitid, rec.RecommendationRating.FAVORITE)
+        assert 1 == self.ratingCount()
+
+    def test_get(self):
+        assert rec.get_user_rating(self.userid, self.submitid) is None
+        rec.set_user_rating(self.userid, self.submitid, rec.RecommendationRating.LIKE)
+        assert rec.get_user_rating(self.userid, self.submitid) is rec.RecommendationRating.LIKE
```
kfkitsune: Missing a blank line at the EOF here.

weasyl/controllers/recommendations.py
```diff
@@ -0,0 +1,16 @@
...
+
+@login_required
+def recommendations_list_(request):
+    # TODO(hyena): This will return roughly 100 items. We don't need that many so cut them off.
+    # TODO(hyena): Handle the case where this returns none due to no favorites.
+    recs = select_list(userid=request.userid, rating=d.get_rating(request.userid))
+
+    page = d.common_page_start(request.userid, options=["recommendations"], title="Recommendations")
+    page.append(d.render("recommendation/recommendations.html", [recs]))
+    return Response(d.common_page_end(request.userid, page))
```
kfkitsune: Missing a blank line at the EOF here.

PR #258, opened 2017-09-06
Move journal content to database

charmander:
Requires `WEASYL_ROOT` to be set for the migration.


charmander:

PR #267, opened 2017-09-23
Default to HTTPS for Tumblr links

charmander:
A redirect is served if the target blog doesn’t support it.


charmander:
It wasn’t supported in the past, so themes were created that relied on insecure resources. HTTPS is opt-in to avoid breaking those themes.

Thank you for the review. =)

Issue #243, opened 2017-05-16
Possible to determine whether an e-mail address has an associated account


charmander:
When signing up or attempting to change e-mail addresses, if an address is in use on Weasyl, the request is denied.

kfkitsune:
I'm going to tentatively say that if we are going for complete plausible deniability over the existence of addresses here, the forgotten password path seems to apply here too. Ignoring the fact that usernames are inherently public, from [weasyl/resetpassword.py#L36](https://github.com/Weasyl/weasyl/blob/master/weasyl/resetpassword.py#L36), where ``user`` is the result of a SQL query retrieving the username and email based on the supplied username:

```
# Check the user's email address against the provided e-mail address,
# raising an exception if there is a mismatch
if email != emailer.normalize_address(user.email):
    raise WeasylError("emailInvalid")
```

In a roundabout way, this could be used the same way: given an email address $EMAIL, and a list of candidate user names $NAMES,

```
for user_name in $NAMES:
    attempt resetting password for user_name with email address $EMAIL
```

It's a tad noisy, potentially, but the same could be said for potentially brute-forcing the process to find out which address belongs to a given account.

A better option might be, over in [weasyl/controllers/user.py#L282](https://github.com/Weasyl/weasyl/blob/master/weasyl/controllers/user.py#L282), saying "**Success!** Provided the supplied information matched our records, information on how to reset your password has been sent to your email address." -- Optionally, provide the email address the user provided on the reset page (" ... to your email address, " + form.email), so they have immediate feedback if they incorrectly entered their email address, since we are essentially making the process a dark grey box (they know they entered their username correctly, but their email could potentially be entered incorrectly (and we don't tell them that)).

Just my two cents, here.

kfkitsune:
>  change e-mail addresses

I've just been thinking about this... and I am not entirely sure if there's a way to not leak the existence of an email here with the code as it is currently written (when viewing this issue as a completely isolated topic). If we get #225 reviewed/merged, then that changes things and we can have plausible deniability since we wouldn't be performing an immediate change of the email address if it was to an otherwise valid email.

As it is currently, if the email is changed, the user will know if the change happened or not because the email address would update in the settings, or if it failed because the email address existed (from the WeasylError throwing). With #225 in the mix, we could claim that we sent the verification email to the provided email, but just not send it out as _we_ know the email address is in use.

charmander:
Correct; this depends on #220.

kfkitsune:
[...]

PR #276, opened 2017-11-19
Plausible deniability for email addresses

kfkitsune:
Prevent the fact that an email address is in-use when attempting to change a user's email account, creating a user account, or when attempting to reset a user's password via the forgotten password mechanism.

Addresses #243.


weasyl/login.py
```diff
@@ -174,27 +174,33 @@ def create(form):
...
-        "birthday": birthday,
-        "unixtime": arrow.now(),
-    })
-
-    # Queue verification email
-    emailer.append([email], None, "Weasyl Account Creation", d.render(
-        "email/verify_account.html", [token, sysname]))
-    d.metric('increment', 'createdusers')
+    # Only attempt to create the account if the email is unused
+    if email_unused:
```
charmander: For user-friendliness (and to resist timing attacks…? okay, that’s not very thorough), I think this should always send an e-mail, just with a different body to the effect of: “This e-mail address was used to sign up with the username …, but you already have a Weasyl account named …. You can \[reset your password] or \[request a username change].”

It would also be nice if this could always send exactly one insert query and rely on constraints to check for duplicate usernames and e-mail addresses, but that’s not how `logincreate` works yet.

(On including the existing username in this message: even though we require both username and e-mail address for a password reset and provide no “forgotten username” feature, I’m sure staff would respond to a request for a username sent from the address that registered it anyway.)

weasyl/resetpassword.py
```diff
@@ -31,27 +31,30 @@ def request(form):
     if not user:
         raise WeasylError("loginRecordMissing")
 
-    # Check the user's email address against the provided e-mail address,
-    # raising an exception if there is a mismatch
+    # Check the user's email address against the provided e-mail address
```
charmander: Following up on the parenthesized bit from the last comment: maybe this form shouldn’t ask for a username at all?

weasyl/login.py
kfkitsune: >  [request a username change]
We don't currently expose the 'premium' function for username aliases to the public, though; would this just be emailing ``support@wea...``? Such as, "If you were attempting to change your username, please send an email from your registered email address for your Weasyl account, ${username}, to support@wea... and request that your username be changed."

charmander: `[request a username change](mailto:support@weasyl.com)` was the idea, but we’ll probably want to have a whole discussion about wording.

weasyl/resetpassword.py
```diff
@@ -13,45 +13,37 @@ def checktoken(token):
...
-        VALUES (%(id)s, %(token)s, %(time)s, %(address)s)
-        ON CONFLICT (userid) DO UPDATE SET
-            token = %(token)s,
-            set_time = %(time)s,
-            address = %(address)s
-    """, id=user.userid, token=token, time=now, address=address)
-
-    # Generate and send an email to the user containing a password reset link
-    emailer.append([email], None, "Weasyl Password Recovery", d.render("email/reset_password.html", [token]))
+        "SELECT userid, email FROM login WHERE email = %(email)s",
```
charmander: `email` doesn’t need to be selected here.

weasyl/profile.py
```diff
@@ -581,20 +583,22 @@ def edit_email_password(userid, username, password, newemail, newemailcheck,
...
-        token = security.generate_key(40)
-        # Store the current token & email, updating them to overwrite a previous attempt if needed
-        d.engine.execute("""
-            INSERT INTO emailverify (userid, email, token, createtimestamp)
-            VALUES (%(userid)s, %(newemail)s, %(token)s, NOW())
-            ON CONFLICT (userid) DO
-              UPDATE SET email = %(newemail)s, token = %(token)s, createtimestamp = NOW()
-        """, userid=userid, newemail=newemail, token=token)
+        # Only actually attempt to change the email if unused; prevent finding out if an email is already registered
+        if email_unused:
```
charmander: This one should probably have an alternate message sent too?

weasyl/profile.py
```diff
@@ -570,7 +570,9 @@ def edit_email_password(userid, username, password, newemail, newemailcheck,
         if newemail != newemailcheck:
             raise WeasylError("emailMismatch")
         elif login.email_exists(newemail):
-            raise WeasylError("emailExists")
+            email_unused = False
```
charmander: The call can be moved down to where it’s needed at this point:

```diff
- if email_unused:
+ if not login.email_exists(newemail):
```

weasyl/login.py
```diff
@@ -174,27 +174,41 @@ def create(form):
...
+        })
+
+        # Queue verification email
+        emailer.append([email], None, "Weasyl Account Creation", d.render(
+            "email/verify_account.html", [token, sysname]))
+        d.metric('increment', 'createdusers')
+    else:
+        # The email address in question is already in use; let the user know this via email
+        #   (perhaps they forgot their username/password)
+        query_username = d.engine.scalar("""
```
charmander: This query can serve as the “unused” check too.

weasyl/login.py
```diff
@@ -173,37 +173,62 @@ def create(form):
...
+        d.metric('increment', 'createdusers')
+    else:
+        # Store a dummy record to support plausible deniability of email addresses
+        # So "reserve" the username, but mark the record invalid, and use the token to satisfy the uniqueness
+        #  constraint for the email field (e.g., if there is already a valid, pending row in the table).
+        d.engine.execute(d.meta.tables["logincreate"].insert(), {
+            "token": token,
+            "username": username,
+            "login_name": sysname,
+            "hashpass": '',
```
charmander: It’s probably not practical to eliminate all timing attacks the way the tables are set up right now, but this one seems particularly easy to measure – bcrypt with our current work factor of 13 takes a few hundred milliseconds. If an attacker times the request, they can determine to some certainty whether the e-mail is registered. I think `passhash(password)` should be run here too to avoid that.

libweasyl/libweasyl/models/tables.py
```diff
@@ -376,6 +377,10 @@ def default_fkey(*args, **kwargs):
     Column('email', String(length=100), nullable=False, unique=True),
     Column('birthday', WeasylTimestampColumn(), nullable=False),
     Column('unixtime', WeasylTimestampColumn(), nullable=False),
+    # Used to determine if a record is invalid for purposes of plausible deniability of email addresses
+    #   AKA, create a logincreate entry if an in-use email address is provided, thus preserving the effect of
+    #   a pending username triggering a username taken error.
+    Column('invalid', Boolean(), nullable=True),
```
charmander: Can this default to `False` instead of being nullable?

templates/email/email_in_use_email_change.html
```diff
@@ -0,0 +1,10 @@
+$def with (username)
+Hello!
+
+You are receiving this message because someone attempted to change their registered account's email address to this address. This email address already has a Weasyl account associated with it by the name of "${username}". Only one Weasyl account may be associated with an email address.
+
+No action is required from you, but we wanted to inform you of this in the event you were attempting to point to Weasyl accounts to the same email address.
```
charmander: “to point **two** Weasyl accounts”?

weasyl/test/login/test_create.py
```diff
@@ -173,12 +167,14 @@ def test_create_fails_if_another_account_has_email_linked_to_their_account():
...
-    form = Bag(username=user_name, password='0123456789', passcheck='0123456789',
+    form = Bag(username="user", password='0123456789', passcheck='0123456789',
                email=email_addr, emailcheck=email_addr,
                day='12', month='12', year=arrow.now().year - 19)
-    with pytest.raises(WeasylError) as err:
-        login.create(form)
-    assert 'emailExists' == err.value.value
+    login.create(form)
+    query = d.engine.scalar("""
+        SELECT username FROM logincreate WHERE username = %(username)s AND invalid IS NOT NULL
```
charmander: There’s a leftover `IS NOT NULL` here.

weasyl/test/login/test_create.py
```diff
@@ -221,9 +217,11 @@ def test_create_fails_if_pending_account_has_same_email():
     form = Bag(username="test", password='0123456789', passcheck='0123456789',
                email=email_addr, emailcheck=email_addr,
                day='12', month='12', year=arrow.now().year - 19)
-    with pytest.raises(WeasylError) as err:
-        login.create(form)
-    assert 'emailExists' == err.value.value
+    login.create(form)
+    query = d.engine.scalar("""
+        SELECT username FROM logincreate WHERE username = %(username)s AND invalid IS NOT NULL
```
charmander: And here.

PR #300, opened 2018-01-05
Make thumbnails visible without JavaScript

charmander:
It’s an easy improvement that makes the site usable.


charmander:

PR #301, opened 2018-01-06
Prevent TypeError when visiting a 2FA signin route with no session.

kfkitsune:
Technically, the POST route doesn't need the change (because it is CSRF protected, and tokens are stored in the session), but the change was made to the GET and POST route for consistency.

Fixes Sentry WEASYL-8M and WEASYLTESTING-7


charmander:
Anything to add to `test/test_two_factor_auth.py`?

kfkitsune:
> Anything to add to `test/test_two_factor_auth.py`?

For the base 2FA testing suite, @charmander? Not that I can immediately think of. Unless we want to write a webtest suite for it, but for the moment I can't think of anything to add in terms of testing the base Python code. If there are any suggestions, we can certainly add them, though!

weasyl/controllers/user.py
```diff
@@ -114,8 +114,9 @@ def _cleanup_2fa_session():
 def signin_2fa_auth_get_(request):
     sess = define.get_weasyl_session()
 
-    # Only render page if the password has been authenticated (we have a UserID stored in the session)
-    if '2fa_pwd_auth_userid' not in sess.additional_data:
+    # Only render page if the session exists //and// the password has
+    # been authenticated (we have a UserID stored in the session)
+    if sess and '2fa_pwd_auth_userid' not in sess.additional_data:
```
charmander: Looking back – should this be `or`?

charmander: *`not sess or`

weasyl/controllers/user.py
```diff
@@ -116,7 +116,7 @@ def signin_2fa_auth_get_(request):
 
     # Only render page if the session exists //and// the password has
     # been authenticated (we have a UserID stored in the session)
-    if sess and '2fa_pwd_auth_userid' not in sess.additional_data:
+    if not sess or not sess.additional_data or '2fa_pwd_auth_userid' not in sess.additional_data:
```
charmander: It looks like `sess` can never be `None` – does this just need to be `not sess.additional_data`?

Issue #273, opened 2017-10-18
oauth2 consumer login path does not respect 2FA


skylerbunny:
If I attempt to log into a server which consumes Weasyl's oauth2, I am only prompted for a Weasyl username/password (if I am not already logged in). This is to say that implicitly, I can bypass my own 2FA preference if enabled, logging in via oauth2.

Both standard logins and oauth2 logins should follow the same codepath to validate a user.

PR #304, opened 2018-01-12
Redirect to real login page for OAuth authorization

Syfaro:
Instead of having a custom login form for the OAuth authorization page, redirect to the actual login page. This prevents code duplication and removes the questionable looking login form on the OAuth page. #273 made it sound like 2FA was bypassed, but it appears to just give an unknown error. 

Things I'm not sure about: 
* It doesn't show a message about OAuth on the login page. Would it be useful to say please log in to continue authorizing the app or similar? 
* The 'Not me' link just logs the user out, but does not provide a way to get back to the OAuth form. I'm not sure what the best way of handling this is.


charmander (approved):
Thank you for the fix! I’ll follow up with at least one of those UI improvements.

charmander (approved):

Issue #302, opened 2018-01-11
Proposal: Allow multiple accounts per unique email address.


kfkitsune:
- [x] Add email address to ``logincreate``.
- [x] Make a page for admins to view pending account creations in the ``logincreate`` table.
- [ ] Proposal: Allow more than one user account per unique email address.

<details>
  <summary>
    Background Information
  </summary>
As a consequence of #243, we don't currently save the email address provided by the user in the ``logincreate`` record. Per Hendikins:

> Hendikins @Hendikins 22:32
> An example where having the e-mail address and date would be useful for administrative purposes is where a user has requested their account be renamed, but someone has tried registering an account with the desired username.
> As it is some people already exploit case-sensitivity to create multiple accounts with a single address. We'd also be able to determine whether the username change request or registration attempt came first.
> (Since we treat foo@bar.baz and Foo@bar.baz as separate addresses, even though the provider may not)

As it [currently stands](https://github.com/Weasyl/weasyl/blob/f21dc035befc6ef9cfc5223f9e45e806632266af/weasyl/login.py#L203), we don't store the email due to the uniqueness constraint of the ``logincreate.email`` field. What we do insert is the following:

```
d.engine.execute(d.meta.tables["logincreate"].insert(), {
            "token": token,
            "username": username,
            "login_name": sysname,
            "hashpass": passhash(password),
            "email": token,
            "birthday": arrow.now(),
            "unixtime": arrow.now(),
            "invalid": True,
        })
```
So since we satisfy the uniqueness constraint with the token, we can't tie the pending account back to a (potential) user.
</details>

<br />

<details>
  <summary>
    Completed in #305 - Store the email address in ``logincreate`` when an email conflicts with a registered user
  </summary>
**Proposal**: Add a ``logincreate.invalid_emailProvided`` field that is not constrained to be unique, and contains the email as provided during the account creation attempt. This could then be hooked into the admin tool for generating account verification tokens, and displaying information such as, ``The logincreate record for this token is invalid. The email address <email address> is already in use on Weasyl.``
</details>

<br />

<details>
  <summary>
    Completed in #305 - Create a mechanism so admins can view pending account creations
  </summary>
> Hendikins @Hendikins 22:21
> (Somewhat related: It would be very, very useful to be able to look up the particulars of a pending registration)
> 
> Kyra @kfkitsune 22:24
> Such as a simple listing of the entries in the logincreate table (sans bcrypt hash, of course)?
> 
[...]

PR #305, opened 2018-01-16
Add the ability for admins to view accounts pending creation

kfkitsune:
Provides a new admin-restricted page to view, and optionally purge, account creation requests in the ``logincreate`` table. In the event of an invalid ``logincreate`` record, the email wouldn't normally be displayed. So this also adds an ``invalid_email_addr`` field to the ``logincreate`` table to permit storing the originally provided email (since the email field is unique to the table).

Addresses the first and second proposals in #302.

### DIFF
--- a/weasyl/controllers/admin.py
+++ b/weasyl/controllers/admin.py
@@ -133,6 +133,7 @@ def admincontrol_pending_accounts_get_(request):
     query = d.engine.execute("""
         SELECT token, username, email, invalid_email_addr, unixtime
         FROM logincreate
+        ORDER BY username
     """).fetchall()
 
     return Response(d.webpage(request.userid, "admincontrol/pending_accounts.html", [query]))
@@ -147,13 +148,12 @@ def admincontrol_pending_accounts_post_(request):
     :param request: A Pyramid request.
     :return: HTTPSeeOther to /admincontrol/pending_accounts
     """
-    form = request.web_input(logincreatetoken="")
-    print(form.logincreatetoken)
-    if form.token:
+    logincreatetoken = request.POST.get("logincreatetoken")
+    if logincreatetoken:
         d.engine.execute("""
             DELETE FROM logincreate
             WHERE token = %(token)s
-        """, token=form.logincreatetoken)
+        """, token=logincreatetoken)
 
     raise HTTPSeeOther(location="/admincontrol/pending_accounts")
 
--- a/weasyl/controllers/admin.py
+++ b/weasyl/controllers/admin.py
@@ -131,12 +131,17 @@ def admincontrol_pending_accounts_get_(request):
     :return: A Pyramid response with a webpage containing the pending accounts.
     """
     query = d.engine.execute("""
-        SELECT token, username, email, invalid_email_addr, unixtime
+        SELECT token, username, email, invalid, invalid_email_addr, unixtime
         FROM logincreate
         ORDER BY username
     """).fetchall()
 
-    return Response(d.webpage(request.userid, "admincontrol/pending_accounts.html", [query]))
+    return Response(d.webpage(
+        request.userid,
+        "admincontrol/pending_accounts.html",
+        [query],
+        title="Accounts Pending Creation"
+    ))
 
 
 @admin_only
--- a/weasyl/controllers/admin.py
+++ b/weasyl/controllers/admin.py
@@ -123,6 +123,42 @@ def admincontrol_acctverifylink_(request):
 
 
 @admin_only
+def admincontrol_pending_accounts_get_(request):
+    """
+    Retrieve a listing of any active pending accounts in the logincreate table.
+
+    :param request: The Pyramid request object.
+    :return: A Pyramid response with a webpage containing the pending accounts.
+    """
+    query = d.engine.execute("""
+        SELECT token, username, email, invalid_email_addr, unixtime
+        FROM logincreate
+    """).fetchall()
+
+    return Response(d.webpage(request.userid, "admincontrol/pending_accounts.html", [query]))
+
+
+@admin_only
+@token_checked
+def admincontrol_pending_accounts_post_(request):
+    """
+    Purges a specified logincreate record.
+
+    :param request: A Pyramid request.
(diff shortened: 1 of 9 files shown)


weasyl/controllers/admin.py
```diff
@@ -122,6 +122,42 @@ def admincontrol_acctverifylink_(request):
...
+def admincontrol_pending_accounts_get_(request):
+    """
+    Retrieve a listing of any active pending accounts in the logincreate table.
+
+    :param request: The Pyramid request object.
+    :return: A Pyramid response with a webpage containing the pending accounts.
+    """
+    query = d.engine.execute("""
+        SELECT token, username, email, invalid_email_addr, unixtime
+        FROM logincreate
```
charmander: This should probably have an order of some… sort.

weasyl/controllers/admin.py
```diff
@@ -122,6 +122,42 @@ def admincontrol_acctverifylink_(request):
...
+@token_checked
+def admincontrol_pending_accounts_post_(request):
+    """
+    Purges a specified logincreate record.
+
+    :param request: A Pyramid request.
+    :return: HTTPSeeOther to /admincontrol/pending_accounts
+    """
+    form = request.web_input(logincreatetoken="")
+    print(form.logincreatetoken)
```
charmander: Leftover debugging?

weasyl/controllers/admin.py
```diff
@@ -122,6 +122,42 @@ def admincontrol_acctverifylink_(request):
...
+@admin_only
+@token_checked
+def admincontrol_pending_accounts_post_(request):
+    """
+    Purges a specified logincreate record.
+
+    :param request: A Pyramid request.
+    :return: HTTPSeeOther to /admincontrol/pending_accounts
+    """
+    form = request.web_input(logincreatetoken="")
```
charmander:
```python
token = request.POST.get('logincreatetoken')
```

weasyl/cron.py
```diff
@@ -8,7 +8,7 @@
 
 
 def run_periodic_tasks():
-    now = arrow.utcnow()
+    now = arrow.utcnow().timestamp
```
charmander: Instead of this, how about passing `time=now - datetime.timedelta(days=2)` to the query?

kfkitsune: That would still need to be converted to a numeric timestamp, it seems.

```
(tmp-hnzdl9RK) kyra@kitty ~/g/tmp> python
Python 3.5.3 (default, Jan 19 2017, 14:11:04)
[GCC 6.3.0 20170118] on linux
>>> import arrow
>>> import datetime
>>> now = arrow.utcnow()
>>> now - datetime.timedelta(days=2)
<Arrow [2018-01-14T21:47:18.110156+00:00]>
>>> res = now - datetime.timedelta(days=2)
>>> type(res)
<class 'arrow.arrow.Arrow'>
>>> res.timestamp
1515966438
>>>
```

We can use ``timedelta``, if we want but we'd need turn around and issue a ``.timestamp``, otherwise we'd be right back to where we are now with the TypeError against the stored value and what we're setting as the ``time`` parameter.
```
>>> (now - datetime.timedelta(days=2)).timestamp
1515966438
```

Personally, I am fine with either method, but I am not sure if one is better than the other in this situation.

charmander: Oh, `.timestamp` is an int? How are `now.minute` and `now.hour` working below?

kfkitsune: Okay, so I had to dig in to the source for this. It seems it [behaves as if it were](https://github.com/crsmithdev/arrow/blob/master/arrow/arrow.py#L25) a ``datetime`` object, but it implements additional functionality. So the original ``datetime`` can be accessed:

```
>>> now._datetime
datetime.datetime(2018, 1, 17, 0, 27, 48, 796416, tzinfo=tzutc())
>>> now._datetime - datetime.timedelta(days=2)
datetime.datetime(2018, 1, 15, 0, 27, 48, 796416, tzinfo=tzutc())
```

But it also responds to typical ``datetime`` attributes such as ``.hour / .minute``.

```
(tmp-hnzdl9RK) kyra@kitty ~/g/tmp> python
Python 3.5.3 (default, Jan 19 2017, 14:11:04)
[GCC 6.3.0 20170118] on linux
>>> import datetime
>>> import arrow
>>> now = arrow.utcnow()
>>> type(now)
<class 'arrow.arrow.Arrow'>
>>> now.hour
0
>>> now.minute
27
>>> type(now.hour)
<class 'int'>
>>> type(now.minute)
<class 'int'>

>>> today = datetime.datetime.today()
>>> today
datetime.datetime(2018, 1, 16, 19, 36, 6, 186271)
>>> today.minute
36
>>> today.hour
19
>>> type(today)
<class 'datetime.datetime'>
>>> type(now._datetime)
<class 'datetime.datetime'>
```

It looks like it is getting the ``hour``/``minute`` out [via the following code](https://github.com/crsmithdev/arrow/blob/master/arrow/arrow.py#L379):
```
def __getattr__(self, name):
        if name == 'week':
            return self.isocalendar()[1]
        if name == 'quarter':
            return int((self.month-1)/self._MONTHS_PER_QUARTER) + 1
        if not name.startswith('_'):
            value = getattr(self._datetime, name, None)
            if value is not None:
                return value
        return object.__getattribute__(self, name)
```
And I can directly replicate the above by doing:
```
>>> getattr(now._datetime, 'hour', None)
0
>>> getattr(now._datetime, 'minute', None)
27
>>> getattr(now._datetime, 'second', None)
48
>>>
```

So in summary, it looks like it directly passes through the attribute if you do a ``arrow.$NAME``, when the property doesn't begin with an underscore, and gets the attribute from the underlying ``datetime`` object. Having dug at this a bit, this makes sense considering that the project page states that it is a "fully implemented, drop-in replacement for datetime" (and you can do .hour, etc. with datetime).

kfkitsune: (Which then means I should just do a ``now.timestamp`` down in the SQL parameter instead of changing the ``now`` as defined here. Hold on; synchronizing.)

PR #311, opened 2018-01-26
Check for noteid when visiting /note route.

### LINKED ISSUE #248: AttributeError: 'noteid' when visiting ``/note`` route without form data
Visiting ``/note`` as a ``GET`` instead of a ``POST`` fails with ``AttributeError: 'noteid'`` in Sentry, and returns an unhandled error message to the user.

If I am reading the Sentry report right, the error exists because we are trying to use a non-existent key in the ``form`` object.

The fix will be to check if the ``noteid`` is even in form prior to attempting to using it.

Example: https://sentry.weasyldev.com/weasyl/weasyl/issues/1476/events/44447/

kfkitsune:
Fixes #248 / Sentry WEASYL-2S and WEASYLTESTING-8

Visiting the bare ``/note`` route without the ``noteid`` as a query string (such as ``/note?noteid=5``) causes an ``AttributeError``. Prevent this from happening by detecting if ``noteid`` even exists in the ``form`` object from the request, and if not, redirect the user to the ``/notes`` route.

Also adds a few tests as they pertain to notes.


weasyl/test/test_note.py
```diff
@@ -0,0 +1,235 @@
...
+    def test_sending_succeeds_if_only_some_recipients_ignored(self, create_users):
+        # user1 ignores user2/user3; user4 is not ignored
+        test_case_note = _make_test_note(recipient="user2; user3; user4")
+        db_utils.create_ignoreuser(create_users['user1'], create_users['user2'])
+        db_utils.create_ignoreuser(create_users['user1'], create_users['user3'])
+
+        note.send(create_users['user1'], test_case_note)
+        query = d.engine.execute("""
+                    SELECT * FROM message
+                """).fetchall()  # , note_sender=userid, note_recipient=create_users['user1']).first()
```
charmander: Leftover?

weasyl/test/test_note.py
```diff
@@ -0,0 +1,235 @@
...
+        # user1 is ignored by user2/user3; user4 does not ignore user1
+        test_case_note = _make_test_note(recipient="user2; user3; user4")
+        db_utils.create_ignoreuser(create_users['user2'], create_users['user1'])
+        db_utils.create_ignoreuser(create_users['user3'], create_users['user1'])
+
+        note.send(create_users['user1'], test_case_note)
+
+        query = d.engine.execute("""
+            SELECT * FROM message
+        """).fetchall()  # , note_sender=userid, note_recipient=create_users['user1']).first()
```
charmander: "

weasyl/test/test_note.py
```diff
@@ -0,0 +1,231 @@
...
+    payload = {
+        'rating': 10,
+        'sfwrating': 10,
+        'timezone': 'America/Dawson',
+        'notes': set_note_config if set_note_config is not None else "anyone",
+    }
+    app.post(url='/control/editpreferences', params=payload, headers={'Cookie': cookie})
+
+
+@pytest.fixture(autouse=True)
```
charmander: It doesn’t look like this should be `autouse`d – everything that needs it has a `create_users` parameter.

weasyl/test/web/test_notes.py
```diff
@@ -0,0 +1,59 @@
...
+
+        # Get user2's notes
+        resp = app.get('/notes', headers={'Cookie': user2_cookie})
+
+        # Verify we can see the note we just sent manually
+        resp.mustcontain("test_note_title_xyz")
+        html = resp.html.find_all("a", "unread")
+
+        # Get the note ID out of the HTML
+        regex = re.compile("(?:.*noteid=)(\d+)(?:.*)")
```
charmander: Is the `(?:.*)` necessary?

charmander:
I added some notes webtests in #411.

PR #312, opened 2018-01-27
Vagrant: Add required packages to build Pillow>=5.0.0

kfkitsune:
Since merging Pillow 5.0.0, Pillow will fail to build from source during a ``vagrant provision`` with output similar to the following:

https://paste.weasyldev.com/show/As7fqhhIenbAIFLVvcy3/
https://paste.weasyldev.com/show/yxFdoHekq1T0jhHid5Qn/

Adding the packages in the diff permit a build of the package to succeed from a clean ``vagrant destroy/make distclean/vagrant up``. Packages sourced from https://pillow.readthedocs.io/en/latest/installation.html#building-on-linux

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -41,9 +41,7 @@ apt-get -y install \
     ruby-sass nodejs
 
 # Required to get Pillow >= 5.0.0 to build from source (since we've disabled using wheels from PyPI)
-apt-get -y install libtiff5-dev libjpeg62-turbo-dev zlib1g-dev \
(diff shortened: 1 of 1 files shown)


skylerbunny:
Are -all- of these packages needed for compilation? I've checked this against our build engine, and we don't have all of them on there. The ones we do have (of the new list):

```libtiff5-dev libjpeg62-turbo-dev zlib1g-dev libfreetype6-dev liblcms2-dev```

PR #313, opened 2018-02-03
Leverage HTTP/2 Server Push to deliver render-blocking resources

kfkitsune:
Adds a new Pyramid tween (http2_server_push_tween_factory) which contains items which should be pushed/preloaded (AKA, render-blocking resources). The ultimate intent behind this is to reduce the time taken before the first page paint occurs by sending data which is required for page render at the outset of the request.

Currently, the render blockers appear to be the JS/CSS files in the page header/footer, and they have been added to the server push preload list.

It looks as though nginx may not -yet- support HTTP/2 server-push of assets, however Cloudflare does. ~~I am hoping that Cloudflare will pick up the slack here~~. Having had this branch loaded to testing, Cloudflare does indeed pick up that the ``Link`` header is added, and the resources are pushed out successfully, as seen in Chrome's dev tools indicating that the resources were pushed.


weasyl/middleware.py
```diff
@@ -153,6 +153,48 @@ def cleanup(request):
...
+        d.cdnify_url('/static/typeahead.bundle.min.js'),
+        d.cdnify_url('/static/marked.js?' + d.CURRENT_SHA),
+        d.cdnify_url('/static/scripts.js?' + d.CURRENT_SHA),
+    ]
+    js_preload = ", ".join('<' + item + '>; rel=preload; as=script' for item in js_resource_preload_list)
+    return ", ".join([js_preload, css_preload])
+
+
+# Part of the `Link` header that will be set in the `http2_server_push_tween_factory` function, below
+HTTP2_LINK_HEADER_PRELOADS = str(_generate_http2_server_push_headers())
```
charmander: It’s already a `str`, so you don’t need the cast here.

kfkitsune:
> I’d like to check if this ends up pushing too often.

We could set an ephemeral session cookie, I suppose, that determines if the ``Link:`` header should be sent, if we really want to be sure. That being said, [according to Cloudflare](https://www.cloudflare.com/website-optimization/http2/serverpush/), if the local client already has the resource cached, it will ignore the push and close the connection for the pushed resource(s).

kfkitsune:
> We could set an ephemeral session cookie

I'll actually look into how to do this. Might as well just send it once if we have the capability to check for if we have sent it (we do, and we can tell the UA to set the ephemeral cookie), and we wouldn't need to store any data locally (just check for the existence of a wzl-http2push cookie, or something, similar to the SFW cookie).

weasyl/middleware.py
```diff
@@ -175,7 +175,7 @@ def _generate_http2_server_push_headers():
         d.cdnify_url('/static/scripts.js?' + d.CURRENT_SHA),
     ]
     js_preload = ", ".join('<' + item + '>; rel=preload; as=script' for item in js_resource_preload_list)
-    return ", ".join([js_preload, css_preload])
+    return ", ".join([js_preload, css_preload]).encode('latin-1')
```
charmander: How was this a lint error?

kfkitsune: Well, lint in the sense that it was raised by the ``../../../virtualenv/python2.7.14/lib/python2.7/site-packages/webtest/lint.py`` package; I suspect that Nginx/Cloudflare would still process it without issue, but this was more, "let's satisfy Webtest". (I admit, the commit description could have been a bit more descriptive; I'll ``--amend`` this one if you'd like.)

https://travis-ci.org/Weasyl/weasyl/jobs/362430306#L1285

weasyl/middleware.py
```diff
@@ -153,6 +153,48 @@ def cleanup(request):
...
+    css_preload = ", ".join('<' + item + '>; rel=preload; as=style' for item in css_resource_preload_list)
+
+    js_resource_preload_list = [
+        d.cdnify_url('/static/jquery-2.2.4.min.js'),
+        d.cdnify_url('/static/typeahead.bundle.min.js'),
+        d.cdnify_url('/static/marked.js?' + d.CURRENT_SHA),
+        d.cdnify_url('/static/scripts.js?' + d.CURRENT_SHA),
+    ]
+    js_preload = ", ".join('<' + item + '>; rel=preload; as=script' for item in js_resource_preload_list)
+    return ", ".join([js_preload, css_preload]).encode('latin-1')
```
charmander: Time for my usual minor and unimportant change requests! /o\ Since there should never be any non-ASCII in here, can you encode as ASCII instead of Latin-1? Also, `js_preload` and `css_preload` could be lists with `", ".join(js_preload + css_preload)`. (The lists are never going to be empty, so all it does is DRY the comma, but…)

PR #315, opened 2018-02-14
Use a newer vagrant .box file

skylerbunny:
Built a newer Debian .box file based on 9.3 for testing and possible inclusion in the project.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,6 +55,7 @@ guest-install-libweasyl: .vagrant
 	vagrant ssh -c 'cd weasyl && make install-libweasyl'
 
 .vagrant:
+	vagrant plugin install vagrant-vbguest
 	vagrant up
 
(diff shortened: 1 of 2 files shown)


skylerbunny:
I think I figured out why this fails. The 'boxcutter' Virtualbox packer I've been using does not install the 'virtualbox-guest-dkms' (from contrib) and 'kernel-headers-($uname -r)' packages. Although it does install the guest additions, when the VM boots, it uses a systemd service to build the modules if they're not present, and without those packages, it fails.

I'm cutting a new .box file, and once done, it'll be uploaded and will be put into this branch as a new checksum. We'll see if that fares better.

skylerbunny:
Made a few small changes to better assure startup in case of service failures, in the Vagrantfile script.

Issue #124, opened 2016-08-11
Add List of Banned/Suspended Users to Mod Panel


Hendikins:
(Previously filed by Nanuk as GitLab issue 8, Fiz as Redmine issue 609)

The moderator control panel should have a page where moderators can access a list of banned/suspended accounts. This should provide the following information:
-  Username
-  Ban date
-  Username of mod/admin
-  Length / release date or "permanent"

Hendikins:
For reference purposes, per @kfkitsune the following pSQL query will get a list of suspended or banned users:
```
``SELECT userid, login_name, email, settings FROM login WHERE settings ~ 's' OR settings ~ 'b';``
```

kfkitsune:
Resolved by merge of #319.

PR #319, opened 2018-02-22
/admincontrol/finduser Enhancements

kfkitsune:
Enhancements for #124.

Enables /admincontrol/finduser to search for banned and/or suspended users, and for searching for accounts created before, after, or between a range of dates. Providing a username/ID/email is optional with the new features (e.g., you can search for all banned accounts created between X-Y dates).

### DIFF
--- a/weasyl/moderation.py
+++ b/weasyl/moderation.py
@@ -309,15 +309,15 @@ def finduser(userid, form):
         ))
 
     # Filter for banned and/or suspended accounts
-    if form.suspendedorbanned == "banned":
-        q = q.where(lo.c.settings.op('~')('b'))
-    elif form.suspendedorbanned == "suspended":
-        q = q.where(lo.c.settings.op('~')('s'))
-    elif form.suspendedorbanned == "both":
+    if form.excludeactive == "on":
         q = q.where(d.sa.or_(
             lo.c.settings.op('~')('b'),
             lo.c.settings.op('~')('s'),
         ))
+    if form.excludebanned == "on":
+        q = q.where(lo.c.settings.op('!~')('b'))
+    if form.excludesuspended == "on":
+        q = q.where(lo.c.settings.op('!~')('s'))
 
     # Filter for date-time
     if form.dateafter and form.datebefore:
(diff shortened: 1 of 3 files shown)


weasyl/moderation.py
```diff
@@ -304,10 +307,35 @@ def finduser(userid, form):
...
+    elif form.dateafter:
+        q = q.where(pr.c.unixtime.op('>=')(arrow.get(form.dateafter)))
+    elif form.datebefore:
+        q = q.where(pr.c.unixtime.op('<=')(arrow.get(form.datebefore)))
+
+    # Apply any row offset
+    if form.row_offset:
+        q = q.offset(form.row_offset)
+
+    if not form.userid and not form.username and not form.email and not form.dateafter \
```
charmander: Can this be checked for right after the first line of the function?

weasyl/moderation.py
```diff
@@ -304,10 +307,35 @@ def finduser(userid, form):
             lo.c.email.op('~')(form.email),
             lo.c.email.op('ilike')('%%%s%%' % form.email),
         ))
-    else:
+
+    # Filter for banned and/or suspended accounts
+    if form.excludeactive == "on":
+        q = q.where(d.sa.or_(
```
charmander: `~` being a regex match, you can simplify this to `q.where(lo.c.settings.op('~')('[bs]'))`.

weasyl/moderation.py
```diff
@@ -304,10 +307,35 @@ def finduser(userid, form):
...
+    if form.excludebanned == "on":
+        q = q.where(lo.c.settings.op('!~')('b'))
+    if form.excludesuspended == "on":
+        q = q.where(lo.c.settings.op('!~')('s'))
+
+    # Filter for date-time
+    if form.dateafter and form.datebefore:
+        q = q.where(d.sa.between(pr.c.unixtime, arrow.get(form.dateafter), arrow.get(form.datebefore)))
+    elif form.dateafter:
+        q = q.where(pr.c.unixtime.op('>=')(arrow.get(form.dateafter)))
```
charmander: `>=` is available as a Python operator: `q.where(pr.c.unixtime >= arrow.get(form.dateafter))`

Issue #129, opened 2016-08-11
Admin/manageuser improvement: Show last known IP addresses for user


Hendikins:
(Originally filed as Redmine issue 139 by @skylerbunny)

Show last known IP addresses for user. This would be capped by count or number of days (say, 10 and 30 respectively).

kfkitsune:
Resolved in #328, for the most part. Currently displays user sessions and the associated IPs. We don't currently drop the IP after N-days in the session information, but we can consider that in the future.

Issue #128, opened 2016-08-11
Add IP search to admin control panel


Hendikins:
(Previously filed as Redmine issue 138 by @skylerbunny)

Originally suggested that this could be added to modcontrol/finduser

If we were to go down that path it would be preferable that this tool be moved under admincontrol - no moderator level tasks require use of the tool.

kfkitsune:
Resolved in #328.

Issue #145, opened 2016-08-24
Moderator tools should use UTC for everything


hyena:
To reduce ambiguity I suggest that moderator tools should use UTC for all date/time, including both the display in logs and the interpretation of dates entered by a moderator. This way all moderators would see the same thing for e.g. a user's DOB.

This should probably be considered to be blocked on #96 to avoid merge conflicts.

Hendikins:
Thumbs up from mod directors.

charmander:
Every time stamp now includes a time zone (#1221), birth dates are plain dates like they should be (on the frontend, anyway), and filters (account search ) are UTC.

PR #328, opened 2018-03-14
Track IP/UA for sessions. Track IP for user creation.

kfkitsune:
For #128 / #129 (assuming a user doesn't log out of a session; full tracking can be added later, if still desired (likely as ``login.last_ip_addr``, or some other field)).

Adds IP information to the ``login`` record for newly created users.

Adds IP and user agent information to the ``sessions`` record. User agent information is stored in a separate table to avoid unnecessary waste of storage duplicating user agent strings.


libweasyl/libweasyl/models/tables.py
```diff
@@ -688,13 +691,27 @@ def default_fkey(*args, **kwargs):
...
 
 Index('ind_sessions_created_at', sessions.c.created_at)
 Index('ind_sessions_userid', sessions.c.userid)
 
 
+user_agents = Table(
+    'user_agents', metadata,
+    Column('user_agent_id', Integer(), primary_key=True, nullable=False),
+    # SHA-256 should be unique enough... I don't think we need to worry (yet) about collisions)
+    Column('user_agent_sha256', String(length=64), unique=True, nullable=False),
```
charmander: What does this column do?

kfkitsune: I was intending that to act as the ultimate unique identifier for a user agent. Since I sliced ``request.user_agent[0:1024]``, there's a (potential) risk for a similar user-agent string to be stored. So I SHA-256'd the entire string, in the event an odd-ball UA appears, so we can 'capture' the entire string via the hash.

At least, that was the initial theory behind that column. Not sure how many legitimate UAs are above 1024 characters in length, but I was trying to cover the edge case.

kfkitsune: On reflection, I might just drop the column entirely, and we can (in a future PR) integrate Google Analytics to do the capturing of UA information if there are any horribly crazy UA strings.

That being said, if we drop the column, would the following be a good way of pulling the ID out of the ``user_agents`` table without indexing the entire UA string? (I'm not in front of a PSQL instance right now, so I can't directly test it).

```
def get_user_agent_id(ua_string=None):
    """
    Store 
    """
    if not ua_string:
        return None
    else:
        # Store/Retrieve the UA String. Max UA length stored is 1024 characters.
        query = d.engine.scalar("""
            INSERT INTO user_agents (user_agent)
            VALUES (%(user_agent)s)
            ON CONFLICT DO
                UPDATE SET user_agent = %(user_agent)s
                WHERE user_agent = %(user_agent)s
            RETURNING user_agent_id
        """, user_agent=ua_string[0:1024])
        return query
```
Explaining the above:
- We don't know (initially) if the UA is present, so try to insert the portion of the UA which is the most relevant (AKA, the first 1024 characters should contain the most relevant portions).
- In the event that UA is already stored, update the UA to be the currently processed UA. This is because:
- We want to return the user_agent_id always so we can bind it to the ``session`` for the user. However, based on the way I am reading the [documentation](https://www.postgresql.org/docs/9.6/static/sql-update.html), the ``RETURNING`` clause doesn't fire unless a row is actually updated. So we would then have to force the row to be updated (even though nothing is really being changed from the standpoint of stored data) to satisfy PSQL's requirements for that clause.

If, for some reason, we needed more granular resolution for the full user agent string, we could revisit the hash method. But after reflection, the above code snippet is what I am planning on changing the existing PR's ``login.process_useragent()`` to, unless we should for some reason store the SHA hash.

Hendikins:
I and other administrators have started poking around at this on testing, and there's a few bits and pieces that have come up from the our perspective (sorry if any of this sounds a bit nitpicky):

Registration IP:
- Is it practical to treat the registration IP as the first "session" in the table, rather than storing it separately? That way it's not stored indefinitely (it can just drop off the list after it reaches a given size) and a simple chronological IP list works a little better with the moderation workflow.

/admincontrol/finduser
- Even though it's possible to search via IP, the search results don't contain the last known IP or a link to the information. A link to /admincontrol/manageuser and if practical the last known IP in the search results would be a considerable improvement to the moderation workflow.
- It appears that searching by IP returns all accounts with IPs containing the search terms (which can even be just "."). It's probably more useful for us to only return results _commencing_ with the search terms.

/admincontrol/manageuser
- Change "Active Login Sessions" to "Recent Login Sessions"
- The session times appear to be UTC-7 for some reason. Ideally these should be either UTC (so no changes will be required for #145 in the future) or moderator's local time (less ideal, but at least consistent with the other mod tools).
- Use 1 second precision for session times (greater precision isn't useful for moderation)
- The "Last Login" and most recent session times aren't matching for some reason
- "Last Login" is redundant with a list of recent logins, so we can remove it
- General clean up whilst we're here: Registration time also displays seconds since Unix epoch. I know it's been that way since forever, but it's not useful for moderation and we should get rid of it.

kfkitsune:
> Is it practical to treat the registration IP as the first "session" in the table, rather than storing it separately?

There's only one instance where this results in an IP not being recorded for a user: the period immediately after verifying the account creation via the verify link. If you're fine with this, I can pull the logic/column for recording the signup IP address out. Though, I wouldn't be surprised if most services track this. The one case I can think of immediately where this might be useful is if a person creates multiple sleeper accounts all at once.

> A link to /admincontrol/manageuser

Easy, and done. I'm swapping the current link to the profile (i.e., ``/~user``) to the admincontrol/manageuser page, as there is a button to get to the profile from that page.

> searching by IP returns all accounts with IPs containing the search terms

That was from copying the logic from earlier. It used a string format ``%%%s%%`` as part of the substitution. Searching for leading patterns is easy, and was as simple as making the pattern ``%s%%``.

> The session times appear to be UTC-7 for some reason.

The column is stored as ``timestamp without time zone`` (``ArrowColumn()`` in the ORM model, which is something that is defined in ``libweasyl``, and not in SQLAlchemy as DateTime is). I'm not entirely sure how to convert this to ``timestamp with time zone``/``DateTime(timezone=True)`` without losing data, here. I do agree that it _should_ be UTC, though. This was a pre-existing time, so a change would be needed for #145 even in the absence of this merge request.

> Use 1 second precision for session times

Fixed (by passing the existing timestamp through ``arrow.get``, and formatting as ``YYYY-MM-DD HH:mm:ss``)

> The "Last Login" and most recent session times aren't matching for some reason

Times are set in two different places in the code, and are not stored in the same format. ``login.last_login`` is stored as a PSQL ``integer``, while ``sessions.created_at`` is stored as PSQL ``timestamp without time zone``. That said, sure, I remvoed the ``last_login`` data from the page.

weasyl/login.py
```diff
@@ -32,23 +34,52 @@ def signin(userid):
     # set the userid on the session
     sess = d.get_weasyl_session()
     sess.userid = userid
+    sess.ip_address = ip_address
+    sess.user_agent_id = get_user_agent_id(user_agent)
     sess.save = True
 
 
+def get_user_agent_id(ua_string=None):
```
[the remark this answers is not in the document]
kfkitsune: No, it wouldn't throw in this case. I have it force-truncate to 1024 characters [on line 58](https://github.com/kfkitsune/weasyl/blob/c304f18291fc926faa7b4f18893bd298f81acb89/weasyl/login.py#L58) as part of the parameterized SQL query inserting the UA into the table; specifically ``ua_string=ua_string[0:1024]``. 

So where the 4 in ``1024`` of the below is the 1024th character in a string...
```
LongUALongUALongUA[... one eternity of characters later ...]LongUALongUA1024Longer
```
Upon storing that, it becomes:
```
LongUALongUALongUA[... one eternity of characters later ...]LongUALongUA1024
```

I just tested this on Testing, and you should be able to [see the result here](https://testing-old.weasyl.com/admincontrol/manageuser?name=ikani).

PR #339, opened 2018-04-11
Add brief notice for how to change username to account registration page

kfkitsune:
It was requested by Hendikins for a note to be added to the registration page to notify users on how to change their username, as an alternative to signing up for a completely new account.

This adds that, as an unobtrusive little note underneath the username entry box.


kfkitsune:
Going to close this one for now; unsure if it's still an issue.

PR #341, opened 2018-04-11
Implement OGP meta headers

### LINKED ISSUE #326: Implement OGP so Telegram et al provide rich page previews
Telegram can generate previews of links when they're posted to a group chat. Some other apps (whatsapp and so on) can do this too. It's based on the OGP protocol: www.ogp.me

Several other sites already implement this - and although it's a bit of fence-painting, it'd be nice for us to have it too.

kfkitsune:
Resolves #326, at least partially in the form of OGP ``<meta>`` tags. Telegram instant views are a different thing, and should likely be a stand-alone issue/proposal.

Since the existing ``twitter_card`` methods provide us with practically everything we need for OGP, we can leverage that information in creating the OGP headers. If the UA is claiming to be ``Twitterbot``, we can easily also include the information for the Twitter card meta headers.

The OGP headers should work (but there's no easy way to test it). That being said, the core four are included (``title``, ``type``, ``image``, ``url``), and we also add in ``site_name`` and ``description``.

Additionally, as ``image`` is a requirement of OGP, in the event that the targeted item is 18+, we display ``/static/images/logo-mark-light.svg`` to satisfy the spec. I don't know if there's a different and preferred image, but I'll be happy to switch it out if desired.


weasyl/controllers/detail.py
```diff
@@ -24,8 +24,23 @@ def submission_(request):
...
+        # The "og:" prefix is specified in page_start.html, and og:image is required by the OGP spec, so something must be in there.
+        extras['ogp'] = {
+            'title': twit_card['title'],
+            'site_name': "Weasyl",
+            'type': "website",
+            'url': twit_card['url'],
+            'description': twit_card['description'],
+            # >> BUG AVOIDANCE: https://trello.com/c/mBx51jfZ/1285-any-image-link-with-in-it-wont-preview-up-it-wont-show-up-in-embeds-too
+            #    Image URLs with '~' in it will not be displayed by Discord, so replace ~ with the URL encoded char code %7E
+            'image': twit_card['image:src'].replace('~', '%7E') if 'image:src' in twit_card else define.absolutify_url(
```
charmander: Since this is already an absolute path, use `define.cdnify_url`.

weasyl/controllers/profile.py
```diff
@@ -30,9 +30,21 @@ def profile_(request):
...
+            extras['twitter_card'] = twit_card
+            extras['options'] = ['nocache']
+        # The "og:" prefix is specified in page_start.html, and og:image is required by the OGP spec, so something must be in there.
+        extras['ogp'] = {
+            'title': twit_card['title'],
+            'site_name': "Weasyl",
+            'type': "website",
+            'url': twit_card['url'],
+            'description': twit_card['description'],
+            'image': twit_card['image:src'] if 'image:src' in twit_card else define.absolutify_url('/static/images/logo-mark-light.svg'),
```
charmander: "

PR #342, opened 2018-04-15
Move libweasyl requirement list to setup.py

charmander:
pip 10 compatibility.


skylerbunny:
Installing now works using pip 10. Version for libweasyl is shown as 0.0.0.

charmander:
@kfkitsune Replaced. Does it work now?

PR #364, opened 2018-05-29
Use Pillow for default submission thumbnails

charmander:
Halves JPEG thumbnail sizes; fixes longstanding colour profile (?) bug; improves thumbnail quality. Also allows more extreme aspect ratios on images that are smaller than thumbnails.

If this works well, it can be applied to characters, covers, avatars, custom thumbnails, and past thumbnails, saving some amount of disk space and bandwidth.


charmander:
@hyena This part of the implementation uses only the resized first frame of an animation as its thumbnail, matching the existing sanpera behaviour. Extending PIL to animated covers is more involved (https://pillow.readthedocs.io/en/latest/reference/ImageSequence.html, https://pillow.readthedocs.io/en/5.1.x/handbook/image-file-formats.html#gif), but supported.

libweasyl/libweasyl/images_new.py
```diff
@@ -0,0 +1,116 @@
+# encoding: utf-8
+
+"""
+Image manipulation with Pillow.
+"""
```
hyena: Add a todo here to rename this file when the switchover is complete.

weasyl/submission.py
```diff
@@ -217,8 +230,11 @@ def create_visual(userid, submission,
         submitid, 'submission', submit_media_item)
     orm.SubmissionMediaLink.make_or_replace_link(
         submitid, 'cover', cover_media_item)
-    orm.SubmissionMediaLink.make_or_replace_link(
-        submitid, 'thumbnail-generated', thumb_generated_media_item)
+
+    for thumb_generated_media_item in thumb_generated_media_items:
+        orm.SubmissionMediaLink.make_or_replace_link(
```
charmander: If `thumb_generated_media_items` has multiple elements, is this just going to replace the same link repeatedly?

charmander:
Okay! Let’s just do it.

PR #368, opened 2018-06-05
Use ECB exchange rates

charmander:
https://www.ecb.europa.eu/stats/policy_and_exchange_rates/euro_reference_exchange_rates/html/index.en.html

### DIFF
--- a/weasyl/commishinfo.py
+++ b/weasyl/commishinfo.py
@@ -1,11 +1,15 @@
 # encoding: utf-8
 from __future__ import absolute_import, division
 
+import logging
 import re
 import urllib
 from collections import namedtuple
 from decimal import Decimal
 
+from pyramid.threadlocal import get_current_request
+
+from weasyl import config
 from weasyl import define as d
 from weasyl import macro as m
 from weasyl.cache import region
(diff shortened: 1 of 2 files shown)


weasyl/commishinfo.py
```diff
@@ -56,16 +60,48 @@ def parse_currency(target):
...
-    the European Central Bank as its upstream source.
+    Retrieve most recent currency exchange rates from the European Central Bank.
 
-    This value is cached with a 24h expiry period
-
-    :return: see http://fixer.io/
+    This value is cached with a 24h expiry period.
     """
-    return None
+    if not config.config_read_bool('convert_currency'):
```
kfkitsune: Is this written as intended? If I am understanding ``config.config_read_bool`` correctly, it should pull in the boolean name, then read the value inside of the value name. So as is currently in the example config:
```
+# Whether currency exchange rates should be fetched for /marketplace
+convert_currency = false
```
Would that evaluate to ``if not False:``, so ``True``, then? And if it were set to true, wouldn't this evaluate to False? (Assuming, of course, I am not misreading this.)

PR #372, opened 2018-06-16
Cache failures to fetch exchange rates for a shorter time

charmander:
I’m not sure if it’s succeeded since the first day. This should give it more chances.


weasyl/commishinfo.py
```diff
@@ -59,12 +59,12 @@ def parse_currency(target):
     return int(Decimal(digits) * (10 ** CURRENCY_PRECISION))
 
 
-@region.cache_on_arguments(expiration_time=60 * 60 * 24)
-def _fetch_rates_xml():
+@region.cache_on_arguments(expiration_time=60 * 60 * 24, should_cache_fn=bool)
```
kfkitsune: I might be misunderstanding how Dogpile works here with the [``should_cache_fn``](http://dogpilecache.readthedocs.io/en/latest/api.html#dogpile.cache.region.CacheRegion.get_or_create.params.should_cache_fn) parameter here, but it _appears_ as if the parameter will call the passed function, and use that as a determination as to if it should cache (e.g., if the function in the parameter returns True or False).

Since ``bool`` is a built-in Python type, and not a function as Dogpile expects, I don't know how this would actually function.
```python
print(bool)                ===> Evals to <class 'bool'>
print(bool == True)        ===> Evals to False
print(bool == False)       ===> Evals to False
print(bool == None)        ===> Evals to False
```
I just wanted to mention this since something seems a bit fishy, if I am looking at the docs. (Of course, if I am wrong, then that's fine too, and I've learned something new.)

charmander: @kfkitsune Dogpile calls `should_cache_fn`, so it’ll evaluate either `bool(None)` (false) or `bool("some XML")` (true).

PR #374, opened 2018-06-17
Remove unused column `welcome.deleted`

charmander:
Also verified that there are no rows in production where it isn’t `NULL`.


libweasyl/libweasyl/alembic/versions/a4f3f44dba52_remove_unused_column_welcome_deleted.py
```diff
@@ -0,0 +1,22 @@
...
+
+Revision ID: a4f3f44dba52
+Revises: 5491fc6b3d51
+Create Date: 2018-06-17 04:59:21.194645
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = 'a4f3f44dba52'
+down_revision = '5491fc6b3d51'
```
kfkitsune: Still no issue with this PR, but as an FYI, more than a few DB migrations were made since this was submitted; the ``down_revision`` will need to be updated before merging the PR to master.

PR #375, opened 2018-06-17
Align maxlength of “location” profile field with database column

charmander:
Fixes https://sentry.weasyldev.com/weasyl/weasyl/issues/1762/.

### DIFF
--- a/templates/control/edit_profile.html
+++ b/templates/control/edit_profile.html
@@ -28,7 +28,7 @@ $code:
       <input type="text" id="editprofile-gender" class="input" name="gender" value="${userinfo['gender']}" maxlength="100" />
 
       <label for="editprofile-country">Location</label>
-      <input type="text" id="editprofile-country" class="input" name="country" value="${userinfo['country']}" maxlength="100" />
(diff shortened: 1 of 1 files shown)


templates/control/edit_profile.html
```diff
@@ -28,7 +28,7 @@ <h3>Profile Settings</h3>
       <input type="text" id="editprofile-gender" class="input" name="gender" value="${userinfo['gender']}" maxlength="100" />
 
       <label for="editprofile-country">Location</label>
-      <input type="text" id="editprofile-country" class="input" name="country" value="${userinfo['country']}" maxlength="100" />
+      <input type="text" id="editprofile-country" class="input" name="country" value="${userinfo['country']}" maxlength="50" />
```
kfkitsune: FWIW, this will block legitimate uses of the field from generating an error. Unless we're truncating server side (e.g., ``value[0:50]``), someone can still force this to error upon submission to the backend. Granted, the length limits would block it from actually being written out, but... just my two-cents.

charmander: Yep. So far only real users have triggered the error, though; I like to keep errors loud unless they become frequent.

PR #378, opened 2018-06-19
Stateless guest sessions

charmander:
The first cleanup step for the session table.

See the commit summary of 5322eaca0ca86f94ba8ffe157aa0049a661e4d96.

### DIFF
--- a/weasyl/middleware.py
+++ b/weasyl/middleware.py
@@ -131,6 +131,11 @@ def session_tween_factory(handler, registry):
         if sess_obj is None:
             sess_obj = create_guest_session()
 
+            additional_csrf = request.cookies.get('__Host-WZLcsrf')
+
+            if additional_csrf is not None:
+                sess_obj.csrf_tokens.append(additional_csrf)
+
         # BUG: Because of the way our exception handler relies on a weasyl_session, exceptions
         # thrown before this part will not be handled correctly.
         request.weasyl_session = sess_obj
--- a/weasyl/middleware.py
+++ b/weasyl/middleware.py
@@ -1,3 +1,4 @@
+# encoding: utf-8
 from __future__ import absolute_import
 
 import os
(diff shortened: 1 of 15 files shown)


weasyl/middleware.py
```diff
@@ -81,45 +83,65 @@ def session_tween_factory(handler, registry):
...
-        if 'WZL' in request.cookies:
-            sess_obj = session.query(orm.Session).get(request.cookies['WZL'])
-            if sess_obj is None:
-                # clear an invalid session cookie if nothing ends up trying to create a
-                # new one
-                cookies_to_clear.add('WZL')
+        cookie = request.cookies.get('WZL')
+
+        if cookie is not None:
+            if is_guest_token(cookie):
```
kfkitsune: Observation: A peculiar effect of this is the ability to arbitrarily pick one's own ``WZL`` token (e.g., ``12345123451234512345``), and still have a subsequent authentication (a CSRF protected action) succeed. Since we're force-changing the ``WZL`` token upon authentication, it doesn't matter, but I just found that ... slightly amusing (but since we're not saving it anywhere, it makes sense that we have to prime the CSRF token from somewhere).

weasyl/test/web/test_login.py
```diff
@@ -58,11 +55,64 @@ def test_2fa_changes_token():
...
+    d.engine.execute("UPDATE sessions SET csrf_token = %(csrf)s WHERE sessionid = %(id)s", id=old_cookie, csrf=old_csrf)
+
+    resp = app.get('/', headers={'Cookie': 'WZL=' + old_cookie})
+    csrf = resp.html.find('html')['data-csrf-token']
+
+    assert 'WZL' not in app.cookies
+    assert '__Host-WZLcsrf' in app.cookies
+    assert csrf == old_csrf
+
+    app.post('/signin', {'token': old_csrf, 'username': 'user2', 'password': 'password2'})
```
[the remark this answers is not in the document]
charmander: It’s checking that it responds with a success status code (i.e. that the CSRF token is considered valid). A 403 would cause WebTest to throw an exception. The username and password are incorrect because it’s not necessary to test the process of logging in, just that guests can submit forms.

weasyl/middleware.py
charmander: Yeah, it’s a bit awkward and why there’s an `isalnum` check in `is_guest_token` (to prevent self-script-injection from the `'<input type="hidden" name="csrf" value="%s" />' % (user_provided_cookie,)`, which probably doesn’t do anything but has no reason to be allowed). Could be avoided with an HMAC, but that would require more characters; I’m not sure if it’s worth it.

PR #390, opened 2018-06-29
Convert Staff YAML to JSON

kfkitsune:
Because it seems peculiar to have a package for a single YAML file, when Python can ingest JSON natively. So let's ditch ``pyyaml`` as a requirement for libweasyl as a bit of housekeeping, and use what Python provides.

NOTE:
Ops _**will**_ need to convert the staff YAML file to JSON before this can be deployed to either testing or prod. Otherwise no staff will exist, as the changed code depends on the .json file to exist to load in staff.

(Also, the example JSON file was cleaned of some old/erroring IDs; I suspect the removed IDs from the YAML-to-JSON conversion weren't exported with the development DB, so it would KeyError upon trying to visit the ``/staff`` page.)


kfkitsune:
Closing in favor of #399; will be reviewing that one.

PR #391, opened 2018-07-01
Add required upgrade step to Alembic migration

kfkitsune:
The prerequisite to adding the CHECK constraint here is no guest sessions remaining. So actually verify our assumptions that guest sessions don't exist by trying to clear guest sessions prior to adding the constraint.

(Labelled bug because we ran into this glitch trying to deploy https://github.com/Weasyl/weasyl/commit/f432db5be5d90e722365ceb45563617bb1c4f0d2 to testing, leading to a violation of the ``CHECK`` constraint. The actual code as it exists is _technically_ correct, but a straight upgrade isn't possible without a manual step.)


kfkitsune:
Closing; per @charmander, this was always meant to be a manual step.

Issue #415, opened 2018-09-02
Comments on site updates


Hendikins:
Add comment threads to site updates.

Blocks potential removal of forums.

charmander:
Deployed!

PR #425, opened 2018-10-22
Small comment optimizations

charmander:
In preparation for #415.


weasyl/comment.py
```diff
@@ -1,3 +1,4 @@
+# encoding: utf-8
```
kfkitsune: Was any special character added in the file? I could be mistaken, but I only see standard ASCII, and no Unicode that would require this addition.
```suggestion
```

charmander: I used an rsquo here:

```python
# the parent comment isn’t visible to this user
```

kfkitsune: Ah! Gotcha. I didn't see that hiding in there.

PR #465, opened 2019-01-22
Add comments to site updates

charmander:
This is not very nice. Sorry.

#415


kfkitsune:
> Okay, so after killing the line at https://github.com/Weasyl/weasyl/blob/master/Vagrantfile#L82 (`rewrite ^/(.*)/$ /\\$1 permanent;`), that fixes the nginx redirect issue.
> 
> Now that it functions, it appears as if this one does what it says on the tin. Comments can now be added to site updates.

Follow-up from Gitter, but removing that line now makes ``/~wesley/`` 404 versus a "user doesn't exist", as ``/~wesley`` does.

PR #468, opened 2019-01-25
Thumbnail/Image Lazy Loading via Intersection Observer API

kfkitsune:
Uses [Y'all.js](https://github.com/malchata/yall.js) to lazily load images. Currently, only images are lazily loaded. It leverages the (Intersection Observer API)[https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API#Browser_compatibility], though it has fallbacks/shims when it isn't available.

Also adds a simple SVG-based loading throbber, which is CC0 licensed via ``loading.io``. If there's an alternate one desired, it can be replaced, but I was mainly looking for something quick to use, and where there wouldn't be any licensing issues when including in the repository.

There is also a graceful fallback via ``<noscript>`` tags to permit loading of the image when scripts are disabled.

There are configurable settings which can be passed to the call to ``yall``, of which this PR only uses ``threshold`` to specify that the element to be loaded needs to be within the viewport by 100 pixels before loading. The default is 200.


kfkitsune:
- [ ] Note to self: Update ``yall`` to [version 3.1.1](https://github.com/malchata/yall.js/blob/3.1.1/dist/yall.min.mjs). ECMAScript 6 looks to be [supported by](https://kangax.github.io/compat-table/es6/) all current browsers, so the ``mjs`` version should work fine?

templates/common/page_start.html
```diff
@@ -18,6 +18,9 @@
   $else:
     <title>Weasyl</title>
 
+  <!-- If we have JavaScript enabled, remove the no-js class -->
```
charmander: This could be a template comment to save sending some bytes, at least:

```suggestion
  $# If we have JavaScript enabled, remove the no-js class
```

templates/common/page_start.html
```diff
@@ -18,6 +18,9 @@
...
+  <script>document.documentElement.classList.remove("no-js")</script>
```
charmander: If this also adds the `js` class, line 184 of `scripts.js` can be removed:

```js
$('html').removeClass('no-js').addClass('js');
```

templates/common/page_end.html
```diff
@@ -73,6 +73,15 @@ <h2 class="typeface-a color-a">Support</h2>
 
 </div><!-- /page-container -->
 
+<script src="${CDNIFY('/static/scripts/yall.min.js')}"></script>
+<script>
+  document.addEventListener("DOMContentLoaded", function() {
```
charmander: Rather than waiting for `DOMContentLoaded` (since this is the end of the page anyway), this can go below the `<script src="/static/scripts.js?${SHA}"></script>` and call `yall` immediately. It’ll be a little faster.

templates/common/thumbnail_grid_item.html
```diff
@@ -55,7 +55,10 @@
           <img src="${display_url}" alt="" />
```
charmander: This one needs to be updated now too at some point, but it’s not critical.

templates/common/page_start.html
kfkitsune: I don't believe ``yall.js`` adds a ``js`` class, when looking at [the source](https://github.com/malchata/yall.js/blob/master/dist/yall.js), or the [usage information](https://github.com/malchata/yall.js).

This specific line was added per the directive on ['What about users without JavaScript?'](https://github.com/malchata/yall.js#what-about-users-without-javascript).

charmander: Sorry, I mean “if you change it to also add a `js` class”. Right now it removes the `no-js` class, which is half of what line 184 does.

templates/common/thumbnail_grid_item.html
```diff
@@ -52,10 +52,19 @@
       $elif webp_thumb is not None:
         <picture>
           <source type="image/webp" srcset="${webp_thumb['display_url']}" />
```
charmander: The yall.js documentation suggests that this should be `data-srcset`:

```suggestion
           <source type="image/webp" data-srcset="${webp_thumb['display_url']}" />
```

static/scripts.js
```diff
@@ -10,14 +10,6 @@
...
 
-
-    // lazy load visible images
-    function loadLazyStuff() {
-        $('img.lazy').filter(':visible').each(function () {
-            this.src = this.getAttribute('data-src');
-        }).removeClass('lazy');
-    }
-
     function debounce(func, threshold) {
```
charmander: As the bot notes, this function is unused now.

templates/common/page_end.html
charmander: I originally thought this wrapper could be removed, but do we actually need to wait for the thumbnail layout to do its thing? It doesn’t change the structure of the DOM from what I can tell, but it does rearrange things. If so, is `DOMContentLoaded` late enough?

On the other hand, we do want visible images to start loading as soon as possible, so maybe it’s worth calling `yall` right away even if it doesn’t have perfect information on what’s visible. `IntersectionObserver` should take care of the rest. I hope.

kfkitsune: This looks like it is going to be required, I think.

- When requiring that the DOMContentLoaded event fire, the page download on my test runs is around 1.34 MiB when viewing the browse and search page for submissions after a hard-refresh, and scrolling down through the page loads all the images for a total page load of 4.88 MiB for all thumbnails; so 27.45% of total file size loads just from hard-refresh.
- When calling ``yall`` directly (so ``<script>yall({threshold: 100});</script>``), the initial download from hard-refresh to idle is 4.21 MiB, with the remaining lazy elements well down in the page. This is 86.27% of the page.

I'll use a direct call if you want, but based on how my test runs are looking to get the best bang for the buck, that's going to mean waiting for ``DOMContentLoaded`` to fire, at least based on how the current page is being rendered (since a direct call seems to fire when more of the elements are on-screen, thus downloading more off-screen elements, due to the image-layout script resizing everything).

charmander: It turns out nothing is holding space for the images, causing that problem when not waiting for thumbnail layout. Something we should probably fix later.

templates/common/thumbnail_grid_item.html
```diff
@@ -51,11 +51,20 @@
...
         </picture>
+        <noscript>
+          <picture>
+            <source type="image/webp" srcset="${webp_thumb['display_url']}" />
+            <img src="${display_url}" alt="" />
+          </picture>
+        </noscript>
       $else:
-        <img src="${display_url}" alt="" />
+        <img class="lazy" data-src="${display_url}" src="data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg'/>" />
```
charmander: Missing `alt`:
```suggestion
        <img class="lazy" data-src="${display_url}" src="data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg'/>" alt="" />
```

templates/common/thumbnail_grid_item.html
```diff
@@ -51,11 +51,20 @@
...
+        <noscript>
+          <img src="${display_url}" />
```
charmander: "
```suggestion
          <img src="${display_url}" alt="" />
```

PR #505, opened 2019-05-01
Fix race when ignoring a user using the profile page button

charmander:
Performs the action, but causes an error page to be displayed.


charmander:
@kfkitsune Thanks for the thorough review!

PR #510, opened 2019-05-13
Implement rate limiting tween/middleware/route decorator

kfkitsune:
Adds a new Pyramid middleware tween, which leverages the dogpile.cache to track requests which fail (via 4xx responses), and rate limits requests on a configurable cooldown window and failed request count.

Inspired, at least in part, by RoR's Rack::Attack, but I couldn't find any pre-made Pyramid middleware of a similar nature. While not as fully featured as Rack::Attack, it should function for our needs.

The middleware tween increments the failed counts on any response which has a value in the 4xx range.

Rate limiting is achieved by decorating a route in the following manner:
```python
# Implies 120 second rate limit window, and 5 failed attempts
@rate_limit_route()
def route_a():
    pass

# The decorator can be configured as desired
@rate_limit_route(expiration_time=60 * 1, limit=6)
def route_a():
    pass
```

Currently, the following routes are limited:
1) /signin
2) /signin/2fa-auth

### DIFF
--- a/weasyl/controllers/user.py
+++ b/weasyl/controllers/user.py
@@ -66,8 +66,6 @@ def signin_post_(request):
         sess.additional_data['2fa_pwd_auth_timestamp'] = arrow.now().timestamp
         # The userid of the user attempting authentication
         sess.additional_data['2fa_pwd_auth_userid'] = logid
-        # The number of times the user has attempted to authenticate via 2FA
-        sess.additional_data['2fa_pwd_auth_attempts'] = 0
         sess.save = True
         return Response(define.webpage(
             request.userid,
@@ -112,11 +110,11 @@ def _cleanup_2fa_session():
     sess = define.get_weasyl_session()
     del sess.additional_data['2fa_pwd_auth_timestamp']
     del sess.additional_data['2fa_pwd_auth_userid']
-    del sess.additional_data['2fa_pwd_auth_attempts']
     sess.save = True
 
 
 @guest_required
+@rate_limit_route(expiration_time=60 * 1)
 def signin_2fa_auth_get_(request):
     sess = define.get_weasyl_session()
 
@@ -145,13 +143,14 @@ def signin_2fa_auth_get_(request):
 
 @guest_required
 @token_checked
+@rate_limit_route(expiration_time=60 * 1)
 def signin_2fa_auth_post_(request):
     sess = define.get_weasyl_session()
     referer = request.params["referer"] if 'referer' in request.params else "/"
 
     # Only render page if the session exists //and// the password has
     # been authenticated (we have a UserID stored in the session)
-    if not sess.additional_data or '2fa_pwd_auth_userid' not in sess.additional_data:
+    if not sess.additional_data and '2fa_pwd_auth_userid' not in sess.additional_data:
         return Response(define.errorpage(request.userid, errorcode.permission))
     tfa_userid = sess.additional_data['2fa_pwd_auth_userid']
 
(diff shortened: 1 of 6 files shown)


weasyl/controllers/user.py
```diff
@@ -139,12 +143,14 @@ def signin_2fa_auth_get_(request):
...
 @token_checked
+@rate_limit_route(expiration_time=60 * 1)
 def signin_2fa_auth_post_(request):
     sess = define.get_weasyl_session()
+    referer = request.params["referer"] if 'referer' in request.params else "/"
 
     # Only render page if the session exists //and// the password has
     # been authenticated (we have a UserID stored in the session)
-    if not sess.additional_data or '2fa_pwd_auth_userid' not in sess.additional_data:
+    if not sess.additional_data and '2fa_pwd_auth_userid' not in sess.additional_data:
```
charmander: This change doesn’t look right. Note that `not` has a higher precedence than `and`.

weasyl/controllers/user.py
```diff
@@ -139,12 +143,14 @@ def signin_2fa_auth_get_(request):
 
 @guest_required
 @token_checked
+@rate_limit_route(expiration_time=60 * 1)
 def signin_2fa_auth_post_(request):
     sess = define.get_weasyl_session()
+    referer = request.params["referer"] if 'referer' in request.params else "/"
```
charmander: `request.params["referer"] if 'referer' in request.params else "/"` is `request.params.get("referer", "/")`

weasyl/middleware.py
```diff
@@ -241,6 +243,37 @@ def http2_server_push(request):
...
+    When the limit is exceeded, subsequent calls against the rate-limited route will raise WeasylError("rateLimitExceeded"),
+    (HTTP 429), until the cached expiration_time passes.
+    """
+    def limit_check(request):
+        resp = handler(request)
+        # Only process the tween if we have indicated that the route is subject to rate limiting
+        if 'Wzl-RateLimit-Count' in request.headers and 'Wzl-RateLimit-Expiration' in request.headers:
+            rate_limit_expiration = request.headers['Wzl-RateLimit-Expiration']
+            rate_limit_max = request.headers['Wzl-RateLimit-Count']
+            cache_key = "mw_rlimit--{addr}|{route_path}".format(addr=request.client_addr, route_path=request.current_route_path())
```
charmander: Rate limiting by IP address seems kind of doomed. (We allow IPv6 too!)

kfkitsune:
> I don’t know how effective this is. We should probably talk more about the right way to rate limit.

Certainly! I was more proposing this change as a marginally better option than what I had been using for the 2FA limiter--that doesn't block re-auth, then another 5 attempts.

As a note, I do see that Cloudflare does have [some level of rate limiting](https://www.cloudflare.com/rate-limiting/) available. That, however, isn't entirely free; that is, the first 10k good requests are free, each 10k after is $0.05. But with that said, if we ensure we return a correct error code (``3xx`` redirect if successful, or ``403`` for failure, for example), I imagine that could cut down on what Cloudflare considers 'good'. Mind, I believe we currently return an ``HTTP200`` by default for failed authentication, which is why I needed to add the ``status=xxx`` to get this iteration to work.

But by all means, I am open to discussions, either here or in a stand-alone issue, about how we should best rate limit requests on routes that shouldn't be able to just be hammered (e.g., login/2fa/etc.).

kfkitsune:
As a thought, and I'll likely work on this later this weekend, but how about using the ``userid`` of the user (claimed, or otherwise) as the limiting factor, instead of IP addresses in the codebase? My reasoning is as follows:

- Usernames are inherently public in Weasyl (e.g., finding a user can be done via ``/~username``, and the system will state if such user exists).
- We either have the ``userid`` [retrieved from login here](https://github.com/Weasyl/weasyl/blob/master/weasyl/login.py#L101), or in the [middleware via the weasyl_session/API token/OAuth2](https://github.com/Weasyl/weasyl/blob/master/weasyl/middleware.py#L252).
- Consequently, we can limit: a) Attempts by a specific user for a specific route; or b) Attempts for a claimed user session to authenticate the TOTP token.

I'm writing this partly for my own benefit in remembering this, and partly for discussion. So in the current code's implementation, that would be checking ``sess.additional_data['2fa_pwd_auth_userid']`` for a TOTP protected session, and ``request.weasyl_session.userid`` for authenticated sessions, so something akin to:
```python
uid = request.weasyl_session.userid or request.weasyl_session.additional_data.get('2fa_pwd_auth_userid')
if uid is not None:
    # Process rate limit code....
```
In the middleware function would likely work.

As an aside, I've been looking over the Pyramid internals, and while I am currently at a loss for how to correctly use it at this moment, I do like the concept they have of both an [authenticated and unauthenticated](https://docs.pylonsproject.org/projects/pyramid-cookbook/en/latest/auth/custom.html) userid. I suppose ``request.weasyl_session.additional_data`` works, though even so Pyramid does seem to have built-in functionality (such as session factories, which was previously [noted by Hyena](https://github.com/Weasyl/weasyl/blob/master/weasyl/middleware.py#L104)).

That's my current two cents and thoughts.

charmander:
> how about using the `userid` of the user (claimed, or otherwise) as the limiting factor, instead of IP addresses in the codebase?

Yep, this is a good idea for TOTP!

kfkitsune:
Closing this one for the moment... a better idea for this would be to do a rate limit on the _account_ versus the IP address.

For instance, if AccountX fails auth--password or 2fa--more than X times in Y minutes (5 times, one minute (two TOTP cycles, IIRC)), forbid additional logins to the account until the duration expires (and HTTP 429 as appropriate).

PR #524, opened 2019-06-21
Messages journals api endpoint

TheWug:


weasyl-apidocs/source/index.rst
```diff
@@ -155,6 +155,28 @@ Basic endpoints
...
+
+   :query nexttime: If specified, only return submissions with a ``unixtime``
+      less (older) than the ``nexttime``. This is used in pagination.
+
+   This will return at most 100 journals. If *count* is more than 100, at
+   most 100 submissions will be returned.
+
+   The result will be a JSON object with three keys: *journals*, *backtime*,
+   and *nexttime*. *journals* will be a JSON array of basic :ref:`journal
+   objects <journals>`. *backtime* and *nextime* are used in
```
charmander: Small typo:
```suggestion
   objects <journals>`. *backtime* and *nexttime* are used in
```

weasyl/controllers/api.py
```diff
@@ -152,6 +152,25 @@ def tidy_submission(submission):
     return submission
 
 
+def tidy_journal(journal):
```
charmander: I think it’s ideal if this returns a new journal instead of modifying the old one, to help make the fields we expect to produce explicit and also to avoid a new field being accidentally exposed in the future. (It’s a bit unlikely with the state of weasyl-old, but it seems like the safe choice.)

… even though `tidy_submission` mutates. (`get_tidy_journal` might be a good distinction.)

weasyl/message.py
```diff
@@ -69,27 +69,86 @@ def remove_all_submissions(userid, only_before=None):
...
         WHERE
             (we.userid, we.type) = (%(user)s, 1010) AND
             rating <= %(rating)s
-        ORDER BY we.unixtime DESC
-    """, user=userid, rating=d.get_rating(userid))
+            {time_filter}
+        {tags_groupby}
+        ORDER BY we.welcomeid DESC
+        {limit_filter}
+        """.format(
```
charmander: Minor formatting consistency with `select_submissions`:
```suggestion
    """.format(
```

libweasyl/setup.py
```diff
@@ -27,7 +27,7 @@
         'alembic==1.0.10',
         'anyjson==0.3.3',
         'arrow==0.14.2',
-        'bcrypt==3.1.6',
+        'bcrypt==3.1.7',
```
charmander: Something strange is going on here. Rebased onto an older master?

weasyl/controllers/api.py
```diff
@@ -152,6 +152,23 @@ def tidy_submission(submission):
     return submission
 
 
+def get_tidy_journal(journal):
+    new_journal = {x: journal[x] for x in journal if x in ["title", "content", "journalid", "tags"]}
+    new_journal['welcomeid'] = journal['id']
+    new_journal['posted_at'] = d.iso8601(journal['unixtime'])
+    contype = journal['contype'] if 'contype' in journal else None
```
charmander: Don’t journal dicts coming into this function always have a `'contype'` (that’s always 30)? Might be best not to include it at all.

weasyl/controllers/api.py
```diff
@@ -152,6 +152,23 @@ def tidy_submission(submission):
...
+    if contype:
+        new_journal['type'] = m.CONTYPE_PARSABLE_MAP[contype]
+    new_journal['rating'] = ratings.CODE_TO_NAME[journal['rating']]
+    new_journal['owner'] = journal['username']
+    new_journal['owner_login'] = d.get_sysname(journal['username'])
+    journalid = journal['journalid']
+    if journalid > 0:
```
charmander: When is a journalid ≤ 0?

TheWug: After reviewing `get_tidy_journal` and comparing it to `tidy_submission`, I conclude that I was overly defensive when I wrote it.

PR #542, opened 2019-07-19
Allow scripts to continue working when storage is disabled

charmander:
Part bug, part enhancement.

### DIFF
--- a/static/scripts.js
+++ b/static/scripts.js
@@ -1575,6 +1575,12 @@
 
     // Home tabs
     (function () {
+        function logStorageError(error) {
+            try {
+                console.warn(error);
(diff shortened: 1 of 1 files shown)


static/scripts.js
```diff
@@ -1604,10 +1604,18 @@
             calculateThumbnailLayout();
             loadLazyStuff();
 
-            localStorage['home-tab'] = paneId;
+            try {
+                localStorage['home-tab'] = paneId;
+            } catch (error) {}
```
hyena: Should these be console logged or is there no good point?

charmander: Sure, that could help find a problem if a change made it throw in a different way.

PR #582, opened 2019-08-28
Return to non-Weasyl web.py at version 0.39

charmander:
Our fix for request paths ending with newlines was released.

(Archived https://github.com/Weasyl/webpy in anticipation of this.)


charmander:
Yep, we need to change how pypi.weasyldev works in the deployment scripts, too. And change what it serves, since it keeps timing out.

PR #583, opened 2019-08-29
Add Akismet spam filtering

kfkitsune:
Adds the capability for spam filtering via Akismet.

NB: I don't know if this is necessarily the best way of doing this, but in theory it should work. @charmander , if you want to push edits directly to the branch to improve upon what I have, by all means ('allow edits from maintainers is always on for a reason). I suspect the journal function is where the majority of the spam is, and that's why the first iteration is focusing here. (And, well, how does one submit image spam content?)

Items remaining to be done for this iteration:
- [x] Add a way to submit new journals to Akismet as spam (also delete here, as well?)
- [x] Add a way to submit submission content (i.e., text) to Akismet.
- [x] Filter out submissions and journals marked as ``is_spam`` from being returned for display. Likely just editing the various ``select_view`` functions.

Items that would need to be done before a deploy:
- Add an [Akismet key](https://akismet.com/development/) to the ``site.config`` file and enable filtering there. FWIW, the fact that we don't have ads yet means that we may not have to pay for the key.

Things to consider:
- Does Cloudflare have anything we can additionally use to help in this regard, especially if the feature(s) are on a free tier, or otherwise no additional charge? @ikanimew, what Cloudflare tier is Weasyl running on currently? According to the [plan page](https://www.cloudflare.com/plans/#compare-features), ``Comment spam protection`` should be available to all tiers. If you need/want me to look at it, let me know and I can create a Cloudflare account to see what it does myself, if you'd like.

Remaining questions:
- ~~How do we fully clean up from deleted submissions? Do we _want_ to actually _**DELETE**_ submissions versus just hiding the record akin to a normal moderation action?~~ Not dealing with this here. Just going to use the hidden setting for the moment.


weasyl/controllers/moderation.py
```diff
@@ -219,3 +219,201 @@ def modcontrol_copynotetostaffnotes_post_(request):
...
+            comment_content=content,
+        )
+        welcome.submission_insert(userid=userid, submitid=submitid, rating=rating, settings=settings)
+    elif action == "delete":
+        # Delete and purge the submission from the welcome table
+        define.engine.execute("""
+            DELETE FROM submission
+            WHERE submitid = %(id)s AND
+                  is_spam = TRUE
+        """, id=submitid)
```
kfkitsune: The more I think on this, the more I dislike directly deleting ``submission`` records.

weasyl/controllers/moderation.py
```diff
@@ -219,3 +219,201 @@ def modcontrol_copynotetostaffnotes_post_(request):
...
+            comment_content=content,
+        )
+        welcome.journal_insert(userid=userid, journalid=journalid, rating=rating, settings=settings)
+    elif action == "delete":
+        # Delete and purge the journal from the welcome table
+        define.engine.execute("""
+            DELETE FROM journal
+            WHERE journalid = %(id)s AND
+                  is_spam = TRUE
+        """, id=journalid)
```
kfkitsune: In theory, ``journal`` records shouldn't be as touchy in terms of remaining artifacts as ``submission`` records (full size image, jpg/png/webp thumbnails), but I don't like this either.

weasyl/spam_filtering.py
```diff
@@ -0,0 +1,140 @@
...
+def _check_if_filtering_enabled(function):
+    """ Block use of `check()` and `submit()` if filtering is not enabled in site.config.txt """
+    def inner(*args, **kwargs):
+        if not FILTERING_ENABLED:
+            raise WeasylError("SpamFilteringDisabled")
+        return function(*args, **kwargs)
+    return inner
+
+
+def _get_user_agent_from_id(id=None):
```
kfkitsune: We really need a utility module of some description... ``define.py`` feels bloated as it is; ``utils.py``? Where even would this be best put? I'm half certain we convert ID-to-UA string other places, so a single source of truth makes sense, versus multiple defined methods. DRY, and all.

templates/modcontrol/spamqueue/journal.html
```diff
@@ -0,0 +1,40 @@
...
+              <b>Journal Title</b>: ${title}
+              <br /><br />
+              <b>Journal Body</b>:
+              <br /><br />
+              ${content}
+            </td>
+            <td>
+              <form action="/modcontrol/spamqueue/journal" method="post">
+                $:{CSRF()}
+                <button class="button positive" name="directive" value="${journal_id},approve">Approve</button>
```
charmander: Would separating the journal id and action into two different fields make more sense here?

```html
<input type="hidden" name="journalid" value="${journal_id}">
<button class="button positive" name="directive" value="approve">Approve</button>
```

templates/modcontrol/spamqueue/submission.html
```diff
@@ -0,0 +1,40 @@
...
+              <b>Submission Title</b>: ${title}
+              <br /><br />
+              <b>Submission Body</b>:
+              <br /><br />
+              ${content}
+            </td>
+            <td>
+              <form action="/modcontrol/spamqueue/submission" method="post">
+                $:{CSRF()}
+                <button class="button positive" name="directive" value="${submission_id},approve">Approve</button>
```
charmander: "

weasyl/controllers/moderation.py
```diff
@@ -219,3 +219,187 @@ def modcontrol_copynotetostaffnotes_post_(request):
...
+    Only displays items that have the `is_spam` flag set, and are not hidden (which is used as the 'reviewed' flag).
+
+    :param request: The Pyramid request.
+    :return: A rendered page for the journal spam queue.
+    """
+    query = define.engine.execute("""
+        SELECT lo.login_name, jo.journalid, jo.title, jo.content
+        FROM login lo
+        INNER JOIN journal jo USING (userid)
+        WHERE jo.is_spam IS TRUE
```
charmander:
```suggestion
        WHERE jo.is_spam
```

weasyl/controllers/moderation.py
```diff
@@ -219,3 +219,187 @@ def modcontrol_copynotetostaffnotes_post_(request):
...
+    Only displays items that have the `is_spam` flag set, and are not hidden (which is used as the 'reviewed' flag).
+
+    :param request: The Pyramid request.
+    :return: A rendered page for the journal spam queue.
+    """
+    query = define.engine.execute("""
+        SELECT lo.login_name, su.submitid, su.title, su.content
+        FROM login lo
+        INNER JOIN submission su USING (userid)
+        WHERE su.is_spam IS TRUE
```
charmander:
```suggestion
        WHERE su.is_spam
```

weasyl/controllers/moderation.py
```diff
@@ -219,3 +219,187 @@ def modcontrol_copynotetostaffnotes_post_(request):
...
+    Either `submitid` or `journalid` must be present in the request's parameters.
+
+    :param request: The Pyramid request.
+    :subparam request.params['submitid']: If present, the submission's ID number.
+    :subparam request.params['journalid']: If present, the journal's ID number.
+    :return/raises: HTTPSeeOther to /modcontrol/suspenduser.
+    """
+    submitid = request.params.get('submitid')
+    journalid = request.params.get('journalid')
+    if sum([1 for item in [submitid, journalid] if item is not None]) != 1:
```
charmander:
```suggestion
    if (submitid is None) ^ (journalid is None):
```

```suggestion
    if sum(item is not None for item in [submitid, journalid]) != 1:
```

One of these options may or may not be clearer?

weasyl/moderation.py
```diff
@@ -561,21 +561,41 @@ def gallery_blacklisted_tags(userid, otherid):
 
 
 def hidesubmission(submitid):
-    d.execute("UPDATE submission SET settings = settings || 'h' WHERE submitid = %i AND settings !~ 'h'", [submitid])
+    d.engine.execute("UPDATE submission SET settings = settings || 'h' WHERE submitid = %(submitid)s AND settings !~ 'h'", submitid=submitid)
```
charmander: Can you split these unrelated changes out into a separate commit? We can merge them right away.

weasyl/spam_filtering.py
```diff
@@ -0,0 +1,139 @@
...
+    _akismet = Akismet(
+        AKISMET_KEY,
+        blog=AKISMET_BLOG_URL,
+        application_user_agent="weasyl/+({0})".format("https://github.com/weasyl/weasyl")
+    )
+else:
+    FILTERING_ENABLED = False
+
+
+def _check_if_filtering_enabled(function):
```
charmander: This appears to be used only once, and I don’t think it’s clearer than just having an inline

```python
def submit(
    ⋮
):
    if not FILTERING_ENABLED:
        raise WeasylError("SpamFilteringDisabled")

    ⋮
```

weasyl/journal.py
```diff
@@ -252,6 +279,7 @@ def select_latest(userid, rating, otherid=None):
     if otherid:
         statement.append(
             " AND jo.userid = %i AND jo.settings !~ '[%sh]'" % (otherid, "" if frienduser.check(userid, otherid) else "f"))
+    statement.append(" AND is_spam = FALSE ")
```
charmander:
```suggestion
    statement.append(" AND NOT is_spam ")
```

weasyl/controllers/moderation.py
```diff
@@ -219,3 +219,192 @@ def modcontrol_copynotetostaffnotes_post_(request):
...
+            is_ham=True,
+            user_ip=ip_addr,
+            user_agent_id=ua_id,
+            user_id=userid,
+            comment_type="submission",
+            comment_content=content,
+        )
+        welcome.submission_insert(userid=userid, submitid=submitid, rating=rating, settings=settings)
+    elif action == "reject":
+        # Note: We need to explicitly convert to int for the moment, otherwise this call fails with:
```
charmander: This is more the rule than the exception – I wouldn’t expect `submitid` to be represented as a string in the other branch. So for now, how about:

```python
action = request.params['directive']
submitid = int(request.params['submitid'])
```

above, with no need for the `all([action, submitid])` check? At some point we can add something to get required values from `request.params` with a `WeasylError` otherwise, but compared to the usual user- and internet-facing endpoints, this one is less important because mods won’t mess with the parameters so any errors that reach Sentry are more likely to be alerting us to an actual mistake.

weasyl/controllers/moderation.py
```diff
@@ -219,3 +219,192 @@ def modcontrol_copynotetostaffnotes_post_(request):
...
+    :subparam request.params['submitid']: If present, the submission's ID number.
+    :subparam request.params['journalid']: If present, the journal's ID number.
+    :return/raises: HTTPSeeOther to /modcontrol/suspenduser.
+    """
+    submitid = request.params.get('submitid')
+    journalid = request.params.get('journalid')
+
+    # Only one parameter should ever be set
+    if sum(item is not None for item in [submitid, journalid]) != 1:
+        raise WeasylError("Unexpected")
```
charmander: Similar here – ids being converted to `int` as late as possible is weird.

```suggestion
        raise WeasylError("Unexpected")

    submitid = int(submitid)
    journalid = int(journalid)
```

weasyl/controllers/moderation.py
```diff
@@ -219,3 +219,192 @@ def modcontrol_copynotetostaffnotes_post_(request):
...
+        WHERE {pkey_name} = %(pkey_value)s
+    """
+
+    if submitid:
+        # The content_type parameter which will be used to signal to the filtering backend what kind of content this is.
+        content_type = "submission"
+        statement = statement.format(table_name="submission", pkey_name="submitid")
+        record_identifier = submitid
+        welcome.submission_remove(submitid=submitid)
+        moderation.hidesubmission(submitid=int(submitid))
```
charmander:
```suggestion
        moderation.hidesubmission(submitid=submitid)
```

PR #624, opened 2019-10-11
Log exceptions when inserting notifications

charmander:
instead of completely ignoring them. They could be important.


charmander:

PR #629, opened 2019-10-18
WIP: Convert the codebase to Python 3

### LINKED ISSUE #576: Python 3: The path forward
So with currently a little over four months [until January 1, 2020](https://github.com/python/devguide/pull/344), we probably should--at minimum--run a conversion of ``weasyl`` to Python 3 before that time.

This issue is intended as a discussion/tracking area for what we need for a successful conversion. I've poked at this on and off with a blunt ten-meter long pole, and here's what I've seen:

- ``-e https://github.com/webpy/webpy.git@b7ea98b7b528467621858a5dccfa4965e0fdc3a4#egg=web.py`` needs to be supplied in place of [``web.py==0.37+weasyl.2``](https://github.com/Weasyl/weasyl/blob/1d926e04910344cebd73951375b52498783f0261/etc/requirements.txt#L3) -- Our copy is not only archived on GitHub, but it doesn't abide by ``absolute_import``, since out copy does, for example, ``from utils import *``; the repository version does ``from .utils import *``.
- If we supply ``--python=python3`` [here](https://github.com/Weasyl/weasyl/blob/1d926e04910344cebd73951375b52498783f0261/Makefile#L44), and also install ``python3-dev python3-virtualenv``, then the core ``weasyl`` package requirements seem to successfully install, and passes through to the libweasyl build stage.
- ``libweasyl`` ran into issues when the provisioner got to that stage, but I am not certain at this point if I placed the interpreter selector in the correct location.

Things we need to do:
- Convert to Python 3's syntax. Minimally, this might be done through ensuring a working branch is clean, ``distclean``ing or otherwise specifying the correct directory, then running a ``2to3`` cycle on it, and unbreaking from there. This is likely the simplest method, since everything would get converted at once. There might be formatting issues or other nuanced changes that need to be made, but this avoids needing to go file by file.
- Alternatively, work inwards from getting a Vagrant provisioning cycle to complete--including the DB migrations (which may need a pre ``2to3`` cycle run)--then running tests iteratively until we have everything passing. This will likely be a bit tricker on this path, given that one file may pull in many other files (e.g., libweasyl, define, etc.) for any given ``.py`` file.

Things I am unsure of:
- How nginx/Twisted selects the python version to use as the interpreter. Is this just done as a consequence of how the ``venv`` is set up?
- How will testing/production be updated for this? Addition of ensuring py3 is installed via the Ansible playbooks, I'd imagine?

Right now I think my biggest issue in groking the conversion process is not knowing how and where the interpreter is selected or otherwise defined to enable the development environment to run in ``py2`` vs. ``py3``. I _think_ the determinant is however the venv is created, but I am not entirely sure.

kfkitsune:
Resolves #576.

Python 2.7.x is growing long in the tooth, and it going the way of the Dodo bird come the start of 2020. So this conversion was more than overdue.

The codebase was primarily converted via ``2to3``, starting from getting a ``vagrant provision`` cycle to complete correctly, then focusing on ``libweasyl``, and finally ``weasyl`` itself.

The biggest issue from the switchover stems from ``d.execute()``, but it was made semi-stable in the current form. ``d.engine.execute()`` should replace all instances of the former, ideally.

Currently in a WIP status, since this has not had a full shakedown, and I suspect there's gaps in the test suite. That being said, the site can be brought up.

(The code for Travis hasn't yet been updated, so I expect this to blow up fantastically when the CI hits it.)


kfkitsune:
Closing in favor of #937.

PR #631, opened 2019-10-19
Update to new DeviantArt name and link format

charmander:


charmander:
It’s a good change for privacy, since subdomains get sort of tossed around.

PR #654, opened 2019-10-31
Don't raise an error if the Akismet backend connection fails.

kfkitsune:
For example, a temporary name resolution failure; it doesn't make sense to force-drop the entire submitted content if Akismet goes down, so behave as if the submission were ham.

We've had a mess of these beginning today, going by Sentry. Since ``Oct 31, 2019 5:38:28 PM UTC``, 119 events affecting 41 users.

Fixes Sentry ID WEASYL-164.

### DIFF
--- a/weasyl/spam_filtering.py
+++ b/weasyl/spam_filtering.py
@@ -2,7 +2,6 @@ import logging
 import os
 
 from akismet import Akismet, SpamStatus
-from pyramid.threadlocal import get_current_request
 
 from weasyl import config
 from weasyl import define as d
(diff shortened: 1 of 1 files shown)


weasyl/spam_filtering.py
```diff
@@ -76,7 +78,16 @@ def check(
         "is_test": _IS_ENVIRONMENT_TESTING,
     }
     if FILTERING_ENABLED:
-        return _akismet.check(**payload)
+        try:
+            return _akismet.check(**payload)
+        except ConnectionError as errmsg:
+            # Don't fail just because of a connection issue to the Akismet backend.
+            request = get_current_request()
+            request.environ['raven.captureMessage']("""
```
kfkitsune: The only thing I am curious about is if this will generate a single event that stacks, or if since the string content is dynamic, if this will spill out into multiple events. Reference ``WEASYL-9R`` where we track [here](https://github.com/Weasyl/weasyl/blob/a7329677411af5f3296bf58e5e5d8f3580be8858/weasyl/commishinfo.py#L79) that the rates have been updated.

charmander: How about just `request.log_exc()`? We’ll be able to tell the difference.

charmander:
Actually, this might be missing a `from requests.exceptions import ConnectionError`?

PR #661, opened 2019-11-04
Fix up list of ignored users

charmander:
- make it not cut off at 20 users,
- sort it case-insensitively,
- show when it’s empty,
- and clean up its code.

### DIFF
--- a/weasyl/ignoreuser.py
+++ b/weasyl/ignoreuser.py
@@ -28,17 +28,12 @@ def cached_list_ignoring(userid):
                      [userid], options=["within"])
 
 
-def select(userid, limit, backid=None, nextid=None):
+def select(userid):
     statement = ["SELECT iu.otherid, pr.username FROM ignoreuser iu"
                  " INNER JOIN profile pr ON iu.otherid = pr.userid"
                  " WHERE iu.userid = %i" % (userid,)]
 
-    if backid:
-        statement.append(" AND pr.username < (SELECT username FROM profile WHERE userid = %i)" % (backid,))
-    elif nextid:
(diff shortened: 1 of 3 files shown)


weasyl/ignoreuser.py
```diff
@@ -28,22 +28,19 @@ def cached_list_ignoring(userid):
...
-                 " WHERE iu.userid = %i" % (userid,)]
-
-    if backid:
-        statement.append(" AND pr.username < (SELECT username FROM profile WHERE userid = %i)" % (backid,))
-    elif nextid:
-        statement.append(" AND pr.username > (SELECT username FROM profile WHERE userid = %i)" % (nextid,))
-
-    statement.append(" ORDER BY pr.username%s LIMIT %i" % (" DESC" if nextid else "", limit))
+def select(userid):
+    results = d.engine.execute(
```
kfkitsune: More of a stylistic choice here, but don't we typically use the ``"""triple quoted strings"""`` for SQL statements?
```suggestion
    results = d.engine.execute("""
        SELECT iu.otherid, pr.username FROM ignoreuser iu
          INNER JOIN profile pr ON iu.otherid = pr.userid
        WHERE iu.userid = %(user)s
        ORDER BY lower(pr.username)
    """, user=userid)
```

charmander: There’s a lot of both in the codebase. I picked this one to include less unnecessary whitespace in the string, because it’s started to bother me that the indents are inside it :P

kfkitsune: Fair enough; LGTM, otherwise.

PR #664, opened 2019-11-06
Replace all uses of string parameters with define.execute

charmander:
Should help with cleaner Python 3 compatibility. :)


weasyl/commishinfo.py
```diff
@@ -352,100 +357,117 @@ def create_price(userid, price, currency="", settings=""):
...
+        " ON CONFLICT (userid) DO UPDATE SET content = %(content)s",
+        user=userid,
+        content=content,
+    )
 
 
 def remove_class(userid, classid):
-    if not d.execute("SELECT EXISTS (SELECT 0 FROM commishclass WHERE (classid, userid) = (%i, %i))",
-                     [d.get_int(classid), userid], ["bool"]):
+    result = d.engine.execute(
```
kfkitsune: I'm curious, does it breath things if a class and price aren't atomically deleted? I doubt there's much change of a deletion failing since the two deletes are practically right next to each other, but I don't currently know if an isolated commishprice entry would negatively affect anything.

weasyl/comment.py
```diff
@@ -151,11 +154,14 @@ def insert(userid, submitid=None, charid=None, journalid=None, updateid=None, pa
...
             raise WeasylError("submissionRecordMissing")
     else:
-        # Determine otherid
-        otherid = d.execute("SELECT userid FROM %s WHERE %s = %i AND settings !~ 'h'",
-                            ["submission", "submitid", submitid] if submitid else
-                            ["character", "charid", charid] if charid else
-                            ["journal", "journalid", journalid], options="element")
+        # Determine the owner of the target
+        otherid = d.engine.scalar(
+            "SELECT userid FROM %s WHERE %s = %i AND settings !~ 'h'" % (
```
kfkitsune: I am pretty sure the input is all shoved through a ``define.get_int()`` before it gets here, but it feels bizarre doing a string format into here directly like this, since this doesn't have the same string escaping like ``d.execute`` did. There's at least one or two more instances of this elsewhere in this PR (if I recall correctly), but is it worth rewriting to ``%(id)s``/``id=submitid if submitid else charid if charid else journalid``, just for sanity's sake?

Just a thought.

charmander: The real protection is `%i`, which only accepts integers – if `define.get_int` were forgotten, it’d throw. It’s equivalent to what `d.execute` would have done, since `d._sql_escape` didn’t change integers.

weasyl/commishinfo.py
charmander: I don’t think it would break anything, but it’s undesirable. I’ll put this in a transaction.

Issue #470, opened 2019-01-27
Site Preferences: "Maximum Viewable Content Rating in SFW Mode" should not include Explicit


kfkitsune:
So, there is a setting on the [Site Preferences](https://www.weasyl.com/control/editpreferences), named "Maximum Viewable Content Rating in SFW Mode" which controls, currently, if General, Mature, or Explicit rated content will be displayed when SFW mode is enabled on a browser.

Except, it doesn't make sense for Explicit to be in there, since at that point that completely negates the point of SFW mode, given that Mature is defined as 18+, but non-sexual in nature.

Perhaps this level should be removed, leaving General and Mature as the only levels available for the ``Maximum Viewable Content Rating in SFW Mode`` dropdown? At least with that change, SFW mode wouldn't be completely negated, and additionally it reduces the possibility for user confusion (since now a user can completely negate the point of SFW mode, and it may seem like the toggle does nothing).

![capture](https://user-images.githubusercontent.com/19718911/51795679-7bb88c00-21b5-11e9-9295-25f443272a97.PNG)

sl1-1:
First time trying to contribute to this project, but I wrote a small javascript function to resolve this issue.
It locks the "Maximum Viewable Content Rating in SFW Mode" drop down to only allow you to select ratings lower than what you have selected as your "Maximum Viewable Content Rating".
Let me know if anything should be changed, or if I should create a PR :)

Issue #187, opened 2016-11-26
User search results have redundant usernames


charmander:
Username  
*by* Username

It might be a good idea to lay results out in columns here with smaller user images, too.

PR #686, opened 2020-01-08
Fixes for issue #470 and #187

sl1-1:
Two small fixes:

1.  Uses a little bit of javascript to prevent people setting the maximum rating in SFW higher than what their maximum rating is normally

2. Removes the extra "By username" when searching users


sl1-1:
Replaced with individual pull requests for each fix.

PR #687, opened 2020-01-08
Switch from anyjson wrapper to python's json

sl1-1:
switching back to the native python json library, instead of anyjson as it seems to be depreciated.


libweasyl/libweasyl/cache.py
```diff
@@ -291,7 +291,7 @@ def save(self, value):
         # turn this into a string because yajl will try to represent this in
```
charmander: This can be removed now, since the built-in json module doesn’t do that.

libweasyl/libweasyl/cache.py
```diff
@@ -239,7 +239,7 @@ def load(self, value):
             return NO_VALUE
         if value.startswith('\0'):
             value = zlib.decompress(value[1:])
-        payload, metadata = anyjson.loads(value)
+        payload, metadata = json.loads(value)
         metadata['ct'] = float(metadata['ct'])
```
charmander: See other comment.

libweasyl/libweasyl/cache.py
sl1-1: might as well clean it up before merging...? that way if we don't move away from json, we aren't left scratching our heads as to why this is still here in a year :)

charmander: True! We do need to remember to clear the cache when this is deployed now.

charmander:
For some reason, I thought simplejson was installed at some point. It isn’t in the now-deployed Docker image, so the anyjson → json part should have no behaviour change. \o/

libweasyl/libweasyl/cache.py
```diff
@@ -288,10 +287,7 @@ def save(self, value):
             :term:`bytes`.
         """
         ret = [value.payload, value.metadata.copy()]
```
charmander: Ah, and this `copy` isn’t necessary anymore either.

PR #688, opened 2020-01-09
Remove Sanpera, Replace with Pillow

sl1-1:
Removes dependency on sanpera, replacing it's functions with pillow


sl1-1:
This is more a looking for feedback pull request, then a do it pull request :)

PR #693, opened 2020-01-26
Remove disk_media table

sl1-1:
Removes the disk_media table, generating the file paths from the SHA256 hash instead of pulling from the table.


sl1-1:
Related to Issue #68

libweasyl/libweasyl/test/common.py
```diff
@@ -66,12 +66,12 @@ def media_item(db, filename, file_type=None):
     The media item created is added to the session.
 
     Returns:
-        Probably a DiskMediaItem.
+        Probably a MediaItem.
```
charmander:
```suggestion
        A MediaItem.
```

It’s certain now!

libweasyl/libweasyl/test/common.py
```diff
@@ -140,7 +140,7 @@ def make_media(db):
     The file used for the media item is fixed: the data file '2x233.gif'.
 
     Returns:
-        Probably a DiskMediaItem.
+        Probably a MediaItem.
```
charmander:
```suggestion
        A MediaItem.
```

libweasyl/libweasyl/models/content.py
```diff
@@ -189,7 +189,7 @@ def set_tags(self, new_tags, done_by_artist=False):
     def create(cls, owner, title, rating, description, category, subtype, folder, tags, friends_only=False,
                critique_requested=False, submission_data=None, cover_data=None, thumbnail_data=None, embed_link=None,
                submission_size_limit=None):
-        from libweasyl.media import fetch_or_create_media_item, make_cover_media_item
+        from libweasyl.media import MediaItem, make_cover_media_item
```
charmander: `MediaItem` should be imported from `libweasyl.models.media`, or `libweasyl.media` should list `'MediaItem'` in its `__all__` where `'fetch_or_create_media_item'` was.

sl1-1: Cleaned this up, and moved a couple of imports that are buried in the code up to the top of the file

libweasyl/libweasyl/models/media.py
```diff
@@ -92,47 +98,30 @@ def make_thumbnail(self, source_image=None):
...
+    @property
+    def file_path(self):
+        path = ['static', 'media'] + fanout(self.sha256, (2, 2, 2)) + ['%s.%s' % (self.sha256, self.file_type)]
+        return os.path.join(*path)
 
-def fetch_or_create_media_item(*a, **kw):
-    return DiskMediaItem.fetch_or_create(*a, **kw)
+    @property
+    def file_url(self):
+        return '/' + self.file_path
```
charmander: For Windows compatibility and a general sense of correctness, there should probably be a third property involved that returns a list of path components. `file_path` can do `os.path.join` and `file_url` can do `'/' + '/'.join`. (Or `full_file_path` could even contain all of `file_path`, since it isn’t used anywhere else.)

PR #699, opened 2020-02-01
Cleanup

charmander:


libweasyl/libweasyl/models/users.py
```diff
@@ -64,60 +59,13 @@ def banner(self):
             return None
         return self.media['banner'][0]
 
-    @reify
```
sl1-1: Wonder if we should keep these?
IMHO, we should probably look to be using the ORM more to validate things.
So instead of in a template doing "userid in staff.MODS" we could do user.is_staff? ( Actually doing that in my jinja2 branch right now)

PR #718, opened 2020-02-08
Remove unused templates

sl1-1:
Remove templates that aren't used by any existing code.


weasyl/templates/user/commissions.html
```diff
@@ -1,29 +0,0 @@
-$def with (profile, userinfo, relationship, query)
-
-<div id="commissions_stage" class="stage">
-  $:{RENDER("common/user_tools.html", [profile, userinfo, relationship])}
-  $:{RENDER("common/user_tabs.html", [profile['username'], None, profile['show_favorites_tab']])}
-</div>
-
-<div id="commissions-content" class="content">
```
charmander: Can you remove the otherwise unused CSS that refers to `#commissions-content` with this?

PR #720, opened 2020-02-08
Fixes #470

### LINKED ISSUE #470: Site Preferences: "Maximum Viewable Content Rating in SFW Mode" should not include Explicit

sl1-1:
Fixes issue #470

### DIFF
--- a/static/scripts.js
+++ b/static/scripts.js
@@ -1714,4 +1714,29 @@
             }
         });
     })();
+
+    (function () {
+        var siterating = $('#siterating');
+        var sfwrating = $('#sfwrating');
+        siterating.change(function (e) {
+            var maxvalue = siterating.children("option:selected").val();
+            $('#sfwrating option').each(function(i, element){
+                if(element.value >= maxvalue && i !== 0){
+                    $(element).hide();
+                }
+                else {
+                    $(element).show();
+                }
+            });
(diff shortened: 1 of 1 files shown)


sl1-1:
> I think the lazy solution is best here, like what it says in the ticket: don’t render the “explicit” option for the SFW rating field. Avoids form weirdness, doesn’t add JavaScript, works without JavaScript.

That won't stop people from being able to do this though
image

charmander:
> That won't stop people from being able to do this though

Correct! It’s not mentioned in the issue but I think we were okay with that when it was first brought up.

PR #722, opened 2020-02-08
Invite mode

charmander:
It’s a bit simplistic, but I hope it’s good enough to let us reopen registration.


charmander:
(I deployed this anyway. :)

PR #733, opened 2020-02-12
Eliminate another round trip for profile.jsonb_settings

charmander:
used on the viewer whenever thumbnails are displayed, for example.

Also fixes a bug: the guest `userid` is `0`, not `None`.


charmander:
Codecov is being its usual buggy self.

PR #739, opened 2020-02-14
Split login CharSettings column to distinct Boolean columns.

### LINKED ISSUE #738: `middleware/status_check_tween_factory` fails for reset birthday/password statuses
The tween effectively catches banned/suspended accounts, sure, but the way this is written, even ``POST``ing to the respective endpoint for reset password/birthday would always cause a redirect to the ``GET`` of the corresponding page, since we're not actually handling the request.

```python
def status_check_tween_factory(handler, registry):
    def status_check_tween(request):
        status = d.common_status_check(request.userid)
        if status:
            # This will always go straight back to the GET! =/
            return Response(d.common_status_page(request.userid, status))
        return handler(request)
    return status_check_tween
```

kfkitsune:
Also fixes a long-standing issue with middleware/status_check_tween_factory, such that a user could never actually reset their password or birthday. This was due to the fact that before the request was handled (i.e., handler(request)), it was always being forcibly redirected via 'return Response(...)'. Resolves #738.


kfkitsune:
I'll be pruning the unused/unexposed birthday reset function. I'll strike out this message once I amend the commit. (Yes, rewriting pushed history is technically bad, but it's not merged to anything yet, and I've seen force pushes elsewhere in merge requests.)

weasyl/define.py
```diff
@@ -319,28 +319,47 @@ def _get_csrf_input():
 
 @region.cache_on_arguments(namespace='v2')
```
charmander: Since the return format is changing, this should be updated to `v3` so it’s not necessary to clear memcached to avoid errors.

PR #741, opened 2020-02-15
Remove define.convert_inputdate in favor of using define.convert_unix…

### LINKED ISSUE #740: Remove and replace `define.convert_inputdate`
It's not user friendly, and HTML5 is a thing with ``<input type="date">``. So why not just directly use ``define.convert_unixdate(day, month, year)``? The HTML5 date picker always returns yyyy/mm/dd, so the format is known.

It's used in two places, as far as I can tell:
1. The forced birthday reset flag page. (Which we may just end up removing anyway; nowhere is it exposed to actually set the 'reset birthday' status to force the screen to appear. And if we're at the point of forcing a birthday reset, an admin knows of it. Why not just do it directly?)
2. ``profile.py:do_manage()``, [L827](https://github.com/Weasyl/weasyl/blob/95fd8a775f3f29ef79e5aa457547cf24e39139b3/weasyl/profile.py#L827), administratively changing a user's birthday. See ``/weasyl/templates/admincontrol/manageuser.html#L72``.

kfkitsune:
…date directly

The function itself used convert_unixdate, so cut out the middleman. Closes #740. Also adds define.convert_iso_date and ISO_DATE (into the template renderer) to translate timestamps to YYYY-MM-DD (ISO 8601) format, which can be loaded into date input fields.

Intentionally ignores the force reset birthday paths, since that will be removed as part of MR #739.

### DIFF
--- a/weasyl/define.py
+++ b/weasyl/define.py
@@ -165,6 +165,7 @@ def _compile(template_name):
                 "CSRF": _get_csrf_input,
                 "USER_TYPE": user_type,
                 "DATE": convert_date,
+                "ISO_DATE": convert_iso_date,
                 "TIME": _convert_time,
                 "LOCAL_ARROW": local_arrow,
                 "PRICE": text_price_amount,
@@ -576,6 +577,20 @@ def convert_date(target=None):
     return result[1:] if result and result[0] == "0" else result
 
 
+def convert_iso_date(target=None):
+    """
+    Converts a unix timestamp to an ISO 8601 date (yyyy-mm-dd). If no target is passed,
+    the current date is returned.
+
+    Note that the date is converted to localtime (via ``convert_to_localtime``).
+
+    :param target: The target unix timestamp to convert. Optional.
+    :return: An ISO 8601 string representing the date of `target`, otherwise the current date formatted as YYYY-MM-DD.
+    """
+    date = convert_to_localtime(target)
+    return arrow.get(date).format("YYYY-MM-DD")
+
+
 def _convert_time(target=None):
     """
     Returns the time in the format 16:00:00. If no target is passed, the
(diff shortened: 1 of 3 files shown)


weasyl/profile.py
```diff
@@ -825,30 +825,33 @@ def do_manage(my_userid, userid, username=None, full_name=None, catchphrase=None
...
-                """,
-                rating_flag=rating_flag,
-                user=userid,
-            )
-            d._get_all_config.invalidate(userid)
-        updates.append('- Birthday: %s' % (birthday,))
+    if birthday is not None:
+        # HTML5 date format is yyyy-mm-dd
+        split = birthday.split("-")
+        if len(split) == 3 and d.convert_unixdate(day=split[2], month=split[1], year=split[0]):
```
charmander: Since the admin form requires an explicit “update” checkbox to update something, I think we should fix the behaviour of ignoring invalid inputs and doing nothing. So:

```python
if len(split) != 3 or d.convert_unixdate(day=split[2], month=split[1], year=split[0]) is None:
    raise WeasylError("birthdayInvalid")
```

weasyl/define.py
```diff
@@ -576,6 +576,20 @@ def convert_date(target=None):
     return result[1:] if result and result[0] == "0" else result
 
 
+def convert_iso_date(target=None):
```
charmander: Since the optional form isn’t used, let’s keep it non-optional for now.

```suggestion
def convert_iso_date(target):
```

weasyl/define.py
```diff
@@ -576,6 +576,20 @@ def convert_date(target=None):
...
+    """
+    Converts a unix timestamp to an ISO 8601 date (yyyy-mm-dd). If no target is passed,
+    the current date is returned.
+
+    Note that the date is converted to localtime (via ``convert_to_localtime``).
+
+    :param target: The target unix timestamp to convert. Optional.
```
charmander:
```suggestion
    :param target: The target unix timestamp to convert.
```

weasyl/define.py
```diff
@@ -576,6 +576,20 @@ def convert_date(target=None):
     return result[1:] if result and result[0] == "0" else result
 
 
+def convert_iso_date(target=None):
+    """
+    Converts a unix timestamp to an ISO 8601 date (yyyy-mm-dd). If no target is passed,
+    the current date is returned.
+
+    Note that the date is converted to localtime (via ``convert_to_localtime``).
```
charmander: It shouldn’t be, right? The original implementation was UTC, and the display/save process shouldn’t depend on the admin’s timezone. (We shouldn’t be storing it as a timestamp at all, but that’s a separate mistake. :)

kfkitsune: I would tend to agree here, though I suppose my reasoning behind this decision was that I was following the same convention as the surrounding functions seeming to use ``convert_to_localtime()``.

Should it instead be something akin to ``date = arrow.get(target) if target else arrow.utcnow()`` and take ``target`` to be a UTC/unix timestamp, instead of just ``date = convert_to_localtime(target)``?

weasyl/profile.py
```diff
@@ -825,30 +825,35 @@ def do_manage(my_userid, userid, username=None, full_name=None, catchphrase=None
...
-
-        if age < ratings.EXPLICIT.minimum_age:
-            max_rating = ratings.GENERAL.code
-            rating_flag = ""
+    if birthday is not None:
+        # HTML5 date format is yyyy-mm-dd
+        split = birthday.split("-")
+        if len(split) != 3 and d.convert_unixdate(day=split[2], month=split[1], year=split[0]) is None:
+            raise WeasylError("birthdayInvalid")
         else:
```
charmander: The `else` can be removed, which also helpfully preserves the original indentation. :)

weasyl/define.py
charmander: `iso8601` uses `datetime.datetime.utcfromtimestamp(unixtime - _UNIXTIME_OFFSET)`. Maybe this could go next to that and be called `iso8601_date`?

weasyl/profile.py
```diff
@@ -825,8 +825,12 @@ def do_manage(my_userid, userid, username=None, full_name=None, catchphrase=None
         updates.append('- Catchphrase: %s' % (catchphrase,))
 
     # Birthday
-    if birthday is not None and d.convert_inputdate(birthday):
-        unixtime = d.convert_inputdate(birthday)
+    if birthday is not None:
+        # HTML5 date format is yyyy-mm-dd
+        split = birthday.split("-")
+        if len(split) != 3 and d.convert_unixdate(day=split[2], month=split[1], year=split[0]) is None:
```
charmander:
```suggestion
        if len(split) != 3 or d.convert_unixdate(day=split[2], month=split[1], year=split[0]) is None:
```

PR #757, opened 2020-02-28
Remove sorttime column from submission table, as it is unused

sl1-1:


sl1-1:
Going to close this, and create a new database cleanup PR

PR #764, opened 2020-03-02
Convert frienduser and watchuser table timestamps

sl1-1:
Converts frienduser and watchuser tables to use DateTime(timezone=True) instead of WeasylTimeColumn. 

Helps out with #68

### DIFF
--- a/libweasyl/libweasyl/alembic/versions/e0320dc462db_convert_frienduser_timestamps_to_.py
+++ b/libweasyl/libweasyl/alembic/versions/e0320dc462db_convert_frienduser_timestamps_to_.py
@@ -8,7 +8,7 @@ Create Date: 2020-02-26 18:18:58.505000
 
 # revision identifiers, used by Alembic.
 revision = 'e0320dc462db'
-down_revision = 'cc2f96b0ba35'
+down_revision = '652d9ba475f0'
 
 from alembic import op   # lgtm[py/unused-import]
 import sqlalchemy as sa  # lgtm[py/unused-import]
--- a/libweasyl/libweasyl/alembic/versions/e0320dc462db_convert_frienduser_timestamps_to_.py
+++ b/libweasyl/libweasyl/alembic/versions/e0320dc462db_convert_frienduser_timestamps_to_.py
@@ -1,0 +1,52 @@
+"""convert frienduser timestamps to DateTime
+
+Revision ID: e0320dc462db
+Revises: 1d8412df7346
+Create Date: 2020-02-26 18:18:58.505000
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = 'e0320dc462db'
+down_revision = 'cc2f96b0ba35'
+
+from alembic import op   # lgtm[py/unused-import]
+import sqlalchemy as sa  # lgtm[py/unused-import]
+
+
+def upgrade():
+    op.alter_column('watchuser', 'unixtime', server_default=None)
+    op.alter_column('watchuser', 'unixtime',
+               existing_type=sa.INTEGER(),
+               server_default=sa.func.now(),
+               type_=sa.DateTime(timezone=True),
+               existing_nullable=False,
+               postgresql_using="timestamp with time zone 'epoch' + (unixtime-18000) * interval '1 second';",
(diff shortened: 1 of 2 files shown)


libweasyl/libweasyl/alembic/versions/e0320dc462db_convert_frienduser_timestamps_to_.py
```diff
@@ -0,0 +1,52 @@
...
+               existing_nullable=False,
+               postgresql_using="timestamp with time zone 'epoch' + (unixtime-18000) * interval '1 second';",
+               new_column_name='created_at')
+    op.alter_column('frienduser', 'unixtime', server_default=None)
+    op.alter_column('frienduser', 'unixtime',
+               existing_type=sa.INTEGER(),
+               server_default=sa.func.now(),
+               type_=sa.DateTime(timezone=True),
+               existing_nullable=False,
+               postgresql_using="timestamp with time zone 'epoch' + (unixtime-18000) * interval '1 second';",
```
charmander: ". Kind of surprised that this works with a semicolon.

```suggestion
               postgresql_using="to_timestamp(unixtime + 18000)",
```

libweasyl/libweasyl/alembic/versions/e0320dc462db_convert_frienduser_timestamps_to_.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+def downgrade():
+    op.alter_column('watchuser', 'created_at', server_default=None)
+    op.alter_column('watchuser', 'created_at',
+                    existing_type=sa.DateTime(timezone=True),
+                    type_=sa.INTEGER(),
+                    existing_nullable=False,
+                    server_default=sa.text(u"(date_part('epoch'::text, now()) - (18000)::double precision)"),
+                    postgresql_using="extract(epoch from created_at)+18000;",
```
charmander:
```suggestion
                    postgresql_using="extract(epoch from created_at) - 18000",
```

libweasyl/libweasyl/alembic/versions/e0320dc462db_convert_frienduser_timestamps_to_.py
```diff
@@ -0,0 +1,52 @@
...
+                    new_column_name='unixtime')
+    op.alter_column('frienduser', 'created_at', server_default=None)
+    op.alter_column('frienduser', 'created_at',
+                    existing_type=sa.DateTime(timezone=True),
+                    type_=sa.INTEGER(),
+                    existing_nullable=False,
+                    server_default=sa.text(u"(date_part('epoch'::text, now()) - (18000)::double precision)"),
+                    postgresql_using="extract(epoch from created_at)+18000;",
```
charmander:
```suggestion
                    postgresql_using="extract(epoch from created_at) - 18000",
```

libweasyl/libweasyl/models/tables.py
```diff
@@ -219,8 +219,7 @@ def default_fkey(*args, **kwargs):
     Column('settings', CharSettingsColumn({
         'p': 'pending',
     }, length=20), nullable=False, server_default='p'),
-    Column('unixtime', WeasylTimestampColumn(), nullable=False,
-           server_default=text(u"(date_part('epoch'::text, now()) - (18000)::double precision)")),
+    Column('created_at', DateTime(timezone=True), nullable=False, server_default=func.now()),
```
charmander: Can you use `TIMESTAMP` instead? It’s not like we’re going to stop using PostgreSQL, but [the documentation][1] does say:

> It is recommended to make use of the `TIMESTAMP` datatype directly when using this flag

(but the PostgreSQL one that’s already imported, not the one it links to.)

I don’t know if this changes the generated migration.


[1]: https://docs.sqlalchemy.org/en/13/core/type_basics.html#sqlalchemy.types.DateTime

sl1-1: Updated, IIRC, they all end up as the same PostgreSQL type, tested it locally, and everything works fine when switched to TIMESTAMP instead of DateTime

PR #767, opened 2020-03-05
Switch to native lazy image loading

### LINKED ISSUE #657: Please add a switch to enable/disable lazyload
This lazyload is very lazy that make user experience very not smooth.
It's better to add a switch go give user more control.

https://github.com/Weasyl/weasyl/issues/639

charmander:
It’s in Firefox Nightly. Fixes #657.


charmander:
It didn’t make it into the HTML standard until pretty recently; I think that’s what they were waiting for.

PR #772, opened 2020-03-08
Convert user timestamps from WeasylTimeStampColumn to TIMESTAMP

sl1-1:


libweasyl/libweasyl/models/tables.py
```diff
@@ -355,8 +355,8 @@ def default_fkey(*args, **kwargs):
     Column('login_name', String(length=40), nullable=False, unique=True),
     Column('hashpass', String(length=100), nullable=False),
     Column('email', String(length=100), nullable=False, unique=True),
-    Column('birthday', WeasylTimestampColumn(), nullable=False),
-    Column('unixtime', WeasylTimestampColumn(), nullable=False),
+    Column('birthday', TIMESTAMP(timezone=True), nullable=False, server_default=func.now()),
```
charmander: I think `birthday` should be a `date`, so we might want to save it for a separate PR.

libweasyl/libweasyl/models/tables.py
```diff
@@ -884,7 +884,7 @@ def default_fkey(*args, **kwargs):
 userinfo = Table(
     'userinfo', metadata,
     Column('userid', Integer(), primary_key=True, nullable=False),
-    Column('birthday', WeasylTimestampColumn(), nullable=False),
+    Column('birthday', TIMESTAMP(timezone=True), nullable=False),
```
charmander: "

weasyl/login.py
```diff
@@ -317,7 +315,7 @@ def verify(token, ip_address=None):
         # Create login record
         userid = db.scalar(lo.insert().returning(lo.c.userid), {
             "login_name": d.get_sysname(query.username),
-            "last_login": arrow.now(),
+            "last_login": datetime.now(tz=pytz.UTC),
```
charmander: Does `sqlalchemy.func.now()` work here?

sl1-1: It needs to be changed around a bit to make it work. Pushed a change with it, let me know if you like it better, or if I should revert it :)

libweasyl/libweasyl/models/tables.py
sl1-1: That almost makes too much sense...
Reverted, will play with it, and make a pull request for that later :)

weasyl/test/login/test_signin.py
```diff
@@ -18,7 +20,8 @@ def test_verify_login_record_is_updated():
     db = d.connect()
     db.add(sess)
     db.flush()
-    d.engine.execute("UPDATE login SET last_login = -1 WHERE userid = %(id)s", id=user_id)
+    time = datetime.fromtimestamp(-1, pytz.UTC)
```
charmander: Ah, we should make this nullable instead of using `1969-12-31T23:59:59Z`.

```python
postgresql_using="CASE last_login WHEN -1 THEN NULL ELSE to_timestamp(last_login + 18000) END"
```
for the upgrade, I guess?

weasyl/test/login/test_verify.py
```diff
@@ -21,7 +21,7 @@ def _create_pending_account(invalid=False):
         "hashpass": login.passhash('0123456789'),
         "email": email,
         "birthday": arrow.Arrow(2000, 1, 1),
-        "unixtime": arrow.now(),
+        "created_at": arrow.now().datetime,
```
charmander: It’s the `server_default`, so does removing it work?

```suggestion
```

weasyl/test/login/test_signin.py
sl1-1: Hmmm, should it be nullable though? We only seem to set it as -1 for this test to make sure it gets updated, the rest of the code doesn't seem to use any timestamps that low.
Theoretically, we could set it to any time prior to when the test is being run?

charmander: Oh, I didn’t realize it was always set in site code. Thanks for explaining!

sl1-1: Made a small change, to make it more clear what we are testing, anyway!

weasyl/test/login/test_verify.py
sl1-1: Yup, makes it a little cleaner!

weasyl/define.py
```diff
@@ -606,7 +608,10 @@ def iso8601_date(target):
     :param target: The target Weasyl timestamp to convert.
     :return: An ISO 8601 string representing the date of `target`.
     """
-    date = datetime.datetime.utcfromtimestamp(target - _UNIXTIME_OFFSET)
+    if isinstance(target, datetime.datetime):
+        date = target
+    else:
+        date = datetime.datetime.utcfromtimestamp(target - _UNIXTIME_OFFSET)
```
charmander: It looks like this is part of the removed birthday change?

```suggestion
    date = datetime.datetime.utcfromtimestamp(target - _UNIXTIME_OFFSET)
```

weasyl/define.py
```diff
@@ -1033,7 +1038,7 @@ def metric(*a, **kw):
 
 
 def iso8601(unixtime):
-    if isinstance(unixtime, arrow.Arrow):
+    if isinstance(unixtime, arrow.Arrow) or isinstance(unixtime, datetime.datetime):
```
charmander: Hm, it’s a bit confusing that `created_at` (good name change!) gets put into fields called `"unixtime"` that can end up here. I think we can sort it out later, though, just by looking at every instance of “unixtime”. (Not a change request :)

weasyl/define.py
sl1-1: Yup!

weasyl/define.py
sl1-1: Ya, I was avoiding trying to make too many template changes with this, so I tried to leave most of the unixtime being passed around. Figure a cleanup sweep would be a good idea after we kill all the unixtime instances?

sl1-1:
> 
> 
> Nice! Now we just have to figure out how to deploy it…

Ya, was trying to think of the best way to do that

sl1-1:
> 
> 
> Nice! Now we just have to figure out how to deploy it…

(Fat fingered while replying!)

Was trying to think of the best way to deploy it. I guess you could manual create the new column, with the defaults, then convert the old values, and then drop the old column? Too bad Alembic didn't handle that part!

PR #808, opened 2020-05-07
Add excerpts and dates to journal list

### LINKED ISSUE #19: Fix gallery view of journals
Presently journals are presented when browsing as tiny user icons with titles that get cut off. This could be done better.

Maybe something more like a list of topics with small user icons next to them?

Design something better.

charmander:
using the same style as the site update list. It’s a bit strange to put an avatar beside each one without a feature of adding different avatars to different journals, but it doesn’t look right to have nothing there either.

| before | after |
|---|---|
| Screenshot_2020-05-07 Argsei's journals — Weasyl | Screenshot_2020-05-07 Argsei's journals — Weasyl |


charmander:
Thank you for the review!

PR #822, opened 2020-06-01
Increase the file size limit for visual submissions from 10 MiB to 50 MiB

charmander:

PR #825, opened 2020-06-10
Update nginx client_max_body_size to 55m

kfkitsune:
Required by #822 (which raised visual submission limit to 50MiB)


charmander:
Thanks for finding that bug!

PR #829, opened 2020-06-22
Fix password reset requests with uppercase in local part

charmander:
The input was normalized with `emailer.normalize_address()`, but not lowercased.


charmander:
Thank you for the review! :)

PR #837, opened 2020-07-10
Remove translationstring

charmander:
We don’t do translations. If we ever start, we won’t miss that rating text was once set up for it.

Also includes removal of unused code.


charmander:
> Doesn't permit setting attributes?

Yep!

PR #840, opened 2020-07-23
profile.resolve cleanup

charmander:


weasyl/profile.py
```diff
@@ -93,7 +93,7 @@ def exchange_settings_from_settings_string(settings_string):
     }
 
 
-def resolve(userid, otherid, othername, myself=True):
+def resolve(userid, otherid, othername):
     """
     Attempts to determine the userid of a specified user; resolves using otherid,
     othername, and userid (if myself is True), in that order. If no userid can be
```
kfkitsune, line 99:
```suggestion
    othername, and userid, in that order. If no userid can be
```
Since we're dropping the ``myself`` parameter.

charmander: Good point :D

PR #797, opened 2020-04-26
remove dependency on storify from controllers/general.py

charmander:
+ adjustments as recommended in code review

Single rebased commit from #744.

PR #744, opened 2020-02-17
Remove storify

sl1-1:
Removes the need for web_input_request_method

PR #852, opened 2020-08-19
Remove dependency on storify from controllers/general.py

sl1-1:
Replaces #797 has requested changes by charmander

### DIFF
--- a/weasyl/controllers/general.py
+++ b/weasyl/controllers/general.py
@@ -83,7 +83,15 @@ def search_(request):
             search.COUNT_LIMIT,
         ]))
     elif find:
-        query = search.browse(request.userid, rating, 66, find, cat, backid, nextid)
+        query = search.browse(
+            userid=request.userid,
+            rating=rating,
+            limit=66,
+            find=find,
+            cat=cat,
+            backid=backid,
+            nextid=nextid,
+        )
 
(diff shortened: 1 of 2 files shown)


weasyl/controllers/general.py
```diff
@@ -77,12 +82,12 @@ def search_(request):
             macro.MACRO_SUBCAT_LIST,
             search.COUNT_LIMIT,
         ]))
-    elif form.find:
-        query = search.browse(request.userid, rating, 66, form)
+    elif find:
+        query = search.browse(request.userid, rating, 66, find, cat, backid, nextid)
```
charmander, line 86: Here too?

```suggestion
        query = search.browse(
            userid=request.userid,
            rating=rating,
            limit=66,
            find=find,
            cat=cat,
            backid=backid,
            nextid=nextid,
        )
```

charmander (changes requested):

weasyl/controllers/general.py
sl1-1: Works for me!

charmander (approved):
Thanks!

sl1-1:
Related to #744 
(just so it shows up on that one, so we can easily find all the related pull requests

PR #853, opened 2020-08-19
Remove storify from controllers/settings.py

sl1-1:
split from #744

### DIFF
--- a/weasyl/controllers/settings.py
+++ b/weasyl/controllers/settings.py
@@ -29,9 +29,11 @@ def control_(request):
 @login_required
 @token_checked
 def control_uploadavatar_(request):
-    form = request.web_input(image="")
+    image = request.params.get('image', u'')
+    if image != u'':
+        image = image.file.read()
 
-    manage = avatar.upload(request.userid, form.image)
+    manage = avatar.upload(request.userid, image)
     if manage:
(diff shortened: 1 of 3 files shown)


weasyl/controllers/settings.py
```diff
@@ -52,34 +54,53 @@ def control_editprofile_get_(request):
...
 
-    if 'more' in form:
-        form.username = define.get_display_name(request.userid)
-        form.sorted_user_links = [(name, [value]) for name, value in zip(form.site_names, form.site_values)]
-        form.settings = form.set_commish + form.set_trade + form.set_request
-        form.config = form.profile_display
-        return Response(define.webpage(request.userid, "control/edit_profile.html", [form, form], title="Edit Profile", options=["typeahead"]))
+    sites = zip(site_names, site_values)
+
+    if 'more' in request.params:
```
kfkitsune, line 68:
```suggestion
    # Allow adding additional links to the profile via 'New' button without intermediate save if JavaScript is disabled
    if 'more' in request.params:
```
The only comment I have here is we might consider a comment here... it took me _way_ too long to grok the code and realize that this ``more`` parameter typically isn't present; it only is if scripts are off, for adding new links without actually saving in-between each newly added entry until done.

PR #855, opened 2020-08-20
Remove dependency on storify from controllers/director.

kfkitsune:
Like it says on the tin. Removes ``web_input`` in favor of Pyramid-native parameter getting mechanisms for the director control panel's controller.

### DIFF
--- a/weasyl/controllers/director.py
+++ b/weasyl/controllers/director.py
@@ -64,6 +64,6 @@ def directorcontrol_globaltagrestrictions_get_(request):
 @director_only
 @token_checked
 def directorcontrol_globaltagrestrictions_post_(request):
-    tags = searchtag.parse_restricted_tags(request.params.get("tags", ''))
+    tags = searchtag.parse_restricted_tags(request.params["tags"])
     searchtag.edit_global_tag_restrictions(request.userid, tags)
     raise HTTPSeeOther(location="/directorcontrol/globaltagrestrictions")
--- a/weasyl/controllers/director.py
+++ b/weasyl/controllers/director.py
@@ -32,19 +32,23 @@ def directorcontrol_emailblacklist_get_(request):
 @token_checked
 @director_only
 def directorcontrol_emailblacklist_post_(request):
-    form = request.web_input(action=None, remove_selection=[], domain_name=None, reason=None)
+    action = request.params.get("action")
+    remove_selection = request.params.getall("remove_selection")
+    domain_name = request.params.get("domain_name")
+    reason = request.params.get("reason")
+
     # Remove entr(y|ies) from blacklist
-    if form.action == "remove":
+    if action == "remove":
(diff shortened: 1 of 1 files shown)


weasyl/controllers/director.py
```diff
@@ -60,6 +64,6 @@ def directorcontrol_globaltagrestrictions_get_(request):
 @director_only
 @token_checked
 def directorcontrol_globaltagrestrictions_post_(request):
-    tags = searchtag.parse_restricted_tags(request.params["tags"])
+    tags = searchtag.parse_restricted_tags(request.params.get("tags", ''))
```
charmander, line 67: What does this change fix?

kfkitsune: Technically nothing in the normal case, and it's likely not needed on this route. I was more thinking, "What if the input isn't on the page entirely?" Probably should just do a bare ``.get("tags")``, then ``if tags: process the tags``. Currently the effect as in this commit is if the input is missing, it purges the restricted tags.

Should we just go for the bare case of a missing input raises an unexpected error? Likely is the most boring solution, but its a toss-up between checking if the parameter is there and just not taking action in that case.

Thoughts?

kfkitsune: So the two options would be, when the ``tags`` input is omitted:
```python
    tags = searchtag.parse_restricted_tags(request.params.get("tags"))  # None, as parameter is missing
    searchtag.edit_global_tag_restrictions(request.userid, tags)  # An unexpected error would raise... somewhere in here.
    raise HTTPSeeOther(location="/directorcontrol/globaltagrestrictions")
```
Versus:
```python
    tags = searchtag.parse_restricted_tags(request.params.get("tags"))  # None, as parameter is missing
    if tags is not None:  # Only process if there is _something_ (even an explicitly blank box), otherwise go back to the tag page
        searchtag.edit_global_tag_restrictions(request.userid, tags)
    raise HTTPSeeOther(location="/directorcontrol/globaltagrestrictions")
```

charmander: I think raising an error is fine. It has the issue of being a reliable way for any client to produce a noisy event via Sentry, but the best way to fix that is by replacing the type of error in some way so it responds with a 4xx status code and an error message that gets sent to the client and isn’t treated as unexpected, instead of ignoring it.

charmander (approved):

weasyl/controllers/director.py
charmander: > It has the issue of being a reliable way for any client to produce a noisy event via Sentry

… or at least that’s usually the issue. This is the director control panel :D

kfkitsune: Fair enough! ``5f39a57`` reverts this line. We'll address further cleanup in the future (but given this is a director route, I really don't anticipate these crazy corner cases popping up often/ever).

charmander:
Merged as f277273!

PR #860, opened 2020-08-23
remove storify from controllers/messages.py

sl1-1:
Related to #744


weasyl/controllers/messages.py
```diff
@@ -14,17 +14,18 @@
...
+    recall = request.params.get('recall', '')
+    remove = request.params.getall('remove')
+    remove_all_before = request.params.get('remove-all-before')
 
     if remove_all_before:
         message.remove_all_before(request.userid, int(remove_all_before))
-    elif form.get('remove-all-submissions'):
-        message.remove_all_submissions(request.userid, define.get_int(form['remove-all-submissions']))
+    elif request.params.get('remove-all-submissions'):
+        message.remove_all_submissions(request.userid, define.get_int(request.params.get('remove-all-submissions')))
```
charmander, line 24: It should be safe to make this stricter by now (it always has an integer value when it’s part of the form; the leniency was for anyone who opened their form before the value for `only_before` was deployed, which was over seven years ago in the weasyl-old commit “Removing submission notifications by date”).

Normally would be changed separately, but if I’m leaving a review comment about `form[` becoming `request.params.get(` anyway… :)

```suggestion
    elif 'remove-all-submissions' in request.params:
        message.remove_all_submissions(request.userid, int(request.params['remove-all-submissions']))
```

charmander (changes requested):

charmander (approved):

Issue #76, opened 2016-06-18
Provisioning fails due to missing gulp-autoprefixer


kfkitsune:
When synced with the current master https://github.com/Weasyl/weasyl/commit/98d0fe8f7aac6d5e9915e593f6c4274f8e0bafb7, building the virtual Weasyl development environment fails.

Output of...
``make setup-vagrant``: https://paste.weasyldev.com/show/1cfNfngydXyCNIq5jXyU/
``make guest-run``: https://paste.weasyldev.com/show/0DyJuVJqmtrHKPLddzWy/

In order to fix this issue and get ``make guest-run`` to work, I had to do the following:

1. ``vagrant ssh``
2. ``cd weasyl``
3. ``npm install --save-dev gulp-autoprefixer`` -- https://www.npmjs.com/package/gulp-autoprefixer
4. ``npm rebuild node-sass`` -- https://github.com/sass/node-sass/issues/1387

After the above, exiting ``vagrant ssh`` and then issuing ``make guest-run`` works as expected.

charmander:
If gulp-autoprefixer was missing, `npm install` probably failed at some earlier point; [it’s in the dependencies][1].

Try removing `node_modules` and running `make guest-run` as usual.

  [1]: https://github.com/Weasyl/weasyl/blob/d5e9f2a99df2b8978dc06a4e0fb9f7450cf6d080/package.json#L4

kfkitsune:
Okay, I have no idea how ``npm install`` failed there, but after ``vagrant destroy``ing, and then redoing ``vagrant up``/``make guest-run``, it finally succeeded as expected.

PR #884, opened 2020-11-04
Add email for signing up while awaiting verification

theSeracen:
This adds another template and calls it when the user attempts to sign
up again with another username, sending another email with the
verification link.

Addresses issue #76

I'm having issues making the dev environment so I wasn't able to test this. I'm trying to fix that and posted a few messages on Gitter, but there wasn't much changed.

### DIFF
--- a/weasyl/login.py
+++ b/weasyl/login.py
@@ -283,9 +283,12 @@ def create(form):
         #   let the already registered user know this via email (perhaps they forgot their username/password)
         query_username_login = d.engine.scalar("SELECT login_name FROM login WHERE email = %(email)s", email=email)
         query_username_logincreate = d.engine.scalar("SELECT login_name FROM logincreate WHERE email = %(email)s", email=email)
-        emailer.send(email, "Weasyl Account Creation - Account Already Exists", d.render(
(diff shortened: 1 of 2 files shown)


weasyl/login.py
```diff
@@ -283,9 +283,12 @@ def create(form):
...
         query_username_logincreate = d.engine.scalar("SELECT login_name FROM logincreate WHERE email = %(email)s", email=email)
-        emailer.send(email, "Weasyl Account Creation - Account Already Exists", d.render(
-            "email/email_in_use_account_creation.html", [query_username_login or query_username_logincreate]))
-
+        if query_username_login:
+            emailer.send(email, "Weasyl Account Creation - Account Already Exists", d.render(
+                "email/email_in_use_account_creation.html", [query_username_login]))
+        else:
+            emailer.send(email, "Weasyl Account Creation - Account Already Exists", d.render(
+                "email/email_in_use_account_creation_in_progress.html", [token, query_username_logincreate]))
```
charmander, line 291: Thanks for the PR – this is a very important feature!

It looks like the token being sent here will correspond to the invalid `logincreate` record created on line 271. Do we need to look up the real one, or did you test it and find out that the one that was intended to be invalid accidentally works in this situation? :D

PR #885, opened 2020-11-07
Docker documentation

theSeracen:
I updated README-docker.md with all of the information from the long, long troubleshooting session to get it working for me :) It now contains all of the steps and information that's required to get the Docker environment up and working. I also included a troubleshooting section for the issues that I ran into, in case they happen again.


kfkitsune (approved):
It works from my perspective. Helped solve a few questions I had, too.

kfkitsune:
Merging as an improvement to the existing Docker readme file. Thanks @theSeracen!

PR #889, opened 2020-11-23
Switch to GitHub Actions for CI

charmander:
Travis CI has been broken on this repository for a while. I don’t really want to go through its permissions mess to fix that. Even when it’s not broken, these days, it can queue for several hours.

Also, GitHub Actions is nicer in a lot of other ways.


charmander:
Well, this doesn’t seem to be running, but it might be because it’s new and from a fork. You can see the run in my fork here: https://github.com/charmander/weasyl/runs/1439695244

requirements/lxml.txt
```diff
@@ -0,0 +1 @@
+lxml==4.5.0
```
kfkitsune, line 1: A few items:
a) I'm slightly confused why we're splitting `lxml` out into its own file!
b) `libweasyl/setup.py:L31` is declaring `'lxml==4.5.2',` is required, but we require `4.5.0` here?

kfkitsune (commented):
A few comments, though I'm not sure if we should pin to an explicit image of Ubuntu versus the `-latest` that is currently in the PR.

Also, yeah, Travis [restructured](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing#building-on-a-public-repositories-only) their pricing plans, with the main stated reason for doing so being some people were abusing the service (e.g., using it to not build/test, and instead, for example, cryptomine) in a significant fashion; which required a change to the existing Travis testing workflows. So that's why the builds started failing. (There are still an allotment of free minutes for open source projects, but that's why it failed.)

.github/workflows/tests.yml
```diff
@@ -0,0 +1,87 @@
+name: Tests
+
+on: [push, pull_request]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
charmander: It seems to be the default, and since we’re running everything in a container inside it anyway (with a similarly specific image python:2.7-alpine3.11, matching production), I think it’s okay.

requirements/lxml.txt
charmander: lxml takes longer to build than the rest of the dependencies combined! So, the Docker-based build [builds it in parallel](https://github.com/Weasyl/weasyl/blob/ec1e99954a846e0f921a30b7b5b64eee6572d3ab/Dockerfile#L22), which is where I got the version number. But apparently I merged some dependabot PRs without thinking about that. Thanks for finding the issue!

kfkitsune (approved):
LGTM!

PR #891, opened 2020-11-23
Make “/” the canonical home page path

charmander:
Most links were this way already, as they should be, but making sure all of them are is cleaner and avoids a duplicate search result.


kfkitsune (approved):
LGTM! No remaining references to `/index` found via `git grep /index`.

charmander:
Thank you for the reviews!

PR #903, opened 2020-11-25
In Markdown preview, avoid running inline event handlers and loading resources before HTML sanitization

### LINKED ISSUE #888: Markdown preview loads resources and runs inline event handlers
> Known vulnerability I discounted when Markdown previews were introduced, completely wrongly because:
>
> - self-XSS is a serious issue
>
> - moderators can edit others’ submission descriptions, with the same instantly-rendering Markdown preview as everywhere else
>
> An upcoming improved fix for this will be eliminating Markdown completely (with any backwards compatibility not including a client-side preview).

This Markdown in a text input with a preview attached to it:

```html
<img src=x onerror=alert(1)>
```

runs the script.

charmander:
Known vulnerability I discounted when Markdown previews were introduced, completely wrongly because:

- self-XSS is a serious issue

- moderators can edit others’ submission descriptions, with the same instantly-rendering Markdown preview as everywhere else

An upcoming improved fix for this will be eliminating Markdown completely (with any backwards compatibility not including a client-side preview).

Fixes #888.

### DIFF
--- a/assets/js/scripts.js
+++ b/assets/js/scripts.js
@@ -710,18 +710,18 @@
             return allowedClasses.indexOf(class_) !== -1;
         }
 
-        function defang(node) {
+        function defang(node, isBody) {
             var i;
 
             for (i = node.childNodes.length; i--;) {
                 var child = node.childNodes[i];
 
                 if (child.nodeType === 1) {
-                    defang(child);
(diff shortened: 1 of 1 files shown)


charmander:
(This is already deployed.)

PR #929, opened 2021-02-20
Remove Flash embedding

charmander:
As of 2021, browsers don’t support it at all.

Continue to allow uploading, since we don’t really offer alternatives yet.


kfkitsune (commented):
NB: Commenting without testing. (Though brief visual review, isn't nixing ``flash.py`` also getting rid of the checking of Flash for correct filetype? So isn't allowing uploads of unchecked files more dangerous under this method? (Even if it requires the user to execute manual steps to download, then execute, the file(s) in question.)).

> Continue to allow uploading, 

I can't say I know how much new--if any--flash content there'd be from this point on, but for a comparison FA seems to have removed the ability to upload Flash (they kept playback via Ruffle). Which leads to...

> since we don’t really offer alternatives yet.

I assume the Nightly variant of [Ruffle](https://ruffle.rs/#usage) via serving the Ruffle [`.js`](https://github.com/ruffle-rs/ruffle/wiki/Using-Ruffle#javascript-api) and [`.wasm`](https://github.com/ruffle-rs/ruffle/wiki/Using-Ruffle#configure-webassembly-mime-type) isn't desired for the moment, for playback, specifically?

I don't quite know how frequently it updates, but I know multiple sites use Ruffle at the moment (e.g., [Newgrounds](https://twitter.com/tomfulp/status/1202210664653185024?lang=en), [FA](https://www.furaffinity.net/journal/9761295/), others). 

[1] https://ruffle.rs/#usage
[2] https://github.com/ruffle-rs/ruffle/wiki/Using-Ruffle#javascript-api
[3] https://github.com/ruffle-rs/ruffle/wiki/Using-Ruffle#configure-webassembly-mime-type

charmander:
Superseded by #935, and Ruffle has a browser extension too.

PR #935, opened 2021-03-02
Adds Ruffle support for SWF playback.

kfkitsune:
Suggested PR over #929 (which removes SWF embeds entirely).

SWF playback via [Ruffle](https://ruffle.rs/).

Note that we'll need to occasionally bump the version via downloading/committing the updates from Ruffle <ruffle.rs>, but this is functional enough where other sites are using this to permit playback of SWF files.

Also note: Nginx will require the wasm type added into the configuration; see /containers/nginx.conf


kfkitsune:
Converting to draft for now since I am still not entirely certain how to handle the resizing of the media viewport to match the ruffle object when loaded.

PR #937, opened 2021-03-04
Python 3

charmander:
- [ ] fix Vagrant setup or make Docker one as usable
- [ ] improve automated test coverage
- [x] do lots of manual testing
- [ ] ~~deploy to separate memcached instance~~


weasyl/config.py
```diff
@@ -1,14 +1,14 @@
 from __future__ import absolute_import
```
kfkitsune, line 1: A preliminary note: The ``__future__`` is now, and not needed on Python 3. (I don't think this _hurts_ anything if it remains until a cleanup PR, but it is an unnecessary relic from Python 2.7.)

PR #942, opened 2021-03-06
Fix coverage configuration

charmander:
- libweasyl’s .coveragerc stopped applying when we stopped running tests from its directory, and its rules should apply to the rest of the project, too.

- Storing .coverage in its own directory makes it possible to mount coverage state onto containers as a volume. (coverage tries to delete the file normally, and creates temporaries around it when using `--cov-append`, so just mounting the file isn’t enough.)


kfkitsune (approved):
Only query, possibly a question for later: Should the existing command to run the tests in Docker in the ``README-docker.md`` execute the coverage checking there? As in, should [pytest-cov](https://pypi.org/project/pytest-cov/) be included just for reference, or is this change mainly intended for getting the GitHub codecov bot to report the coverage?

Otherwise, LGTM, since codecov is reporting correctly.

PR #943, opened 2021-03-06
Markdown cleanup

charmander:
- Adds more tests, bringing up measured line/branch coverage on libweasyl.text to 100% (the “uncovered” `continue` and the associated partial coverage on the `elif` seems to be some kind of coverage bug)
- Simplifies and fixes `` removal

I recommend looking at the diff with whitespace changes ignored.


kfkitsune (approved):
LGTM!

PR #949, opened 2021-03-11
Remove unused `submission.select_*` parameter `exclude`

charmander:
Unused since the beginning of accessible history, 2012-11-19.


kfkitsune (approved):
LGTM via visual review.

PR #950, opened 2021-03-11
Remove unused and mostly unimplemented content-item settings

charmander:
Encores, pools, collaborations, locking. All verified to not be present in the production database, and I didn’t find them anywhere else in the code, although they’re hard to look for because settings columns are terrible.


kfkitsune (approved):
> settings columns are terrible.
Yes. (I suppose they were added in-lieu of additional DB columns?) No instances of singular calls seen (e.g., "t").

LGTM via visual review.

PR #951, opened 2021-03-11
Remove obsolete profile config for tag blocking type

charmander:
`l` indicated use of a tag blocklist, which is the only kind now. ~64k profiles have this applied.

```
=> SELECT round(max(userid), -4) AS max FROM profile;
┌────────┐
│  max   │
├────────┤
│ 190000 │
└────────┘

=> SELECT round(max(userid), -4) AS max FROM profile WHERE config ~ 'l';
┌───────┐
│  max  │
├───────┤
│ 80000 │
└───────┘
```

### DIFF
--- a/etc/SETTINGS.txt
+++ b/etc/SETTINGS.txt
@@ -38,7 +38,6 @@ profile.config (lkscftjone)
 
 b  Show birthday in profile
 
-l  Use only tag blocklist
 a  View max rating mature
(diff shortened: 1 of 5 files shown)


kfkitsune (approved):
LGTM via visual review. Couldn't find any instances remaining using ``"l"`` or ``'l'``. (Streaming excluded, but that isn't this.)

PR #957, opened 2021-03-20
Post-Python-3 package changes

charmander:
See commits!

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,64 +6,24 @@ updates:
     interval: daily
   open-pull-requests-limit: 10
   ignore:
-  - dependency-name: pillow
-    versions:
-    - ">= 7.a"
-    - "< 8"
   - dependency-name: publicsuffixlist
     versions:
     - "> 0.6.13"
-  - dependency-name: pyotp
-    versions:
-    - "> 2.3.0"
   - dependency-name: qrcodegen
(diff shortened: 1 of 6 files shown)


.github/dependabot.yml
```diff
@@ -6,67 +6,24 @@ updates:
...
   open-pull-requests-limit: 10
   ignore:
-  - dependency-name: mock
-    versions:
-    - "> 2.0.0"
-  - dependency-name: pillow
-    versions:
-    - ">= 7.a"
-    - "< 8"
   - dependency-name: publicsuffixlist
```
kfkitsune, line 9: Just curious, why are we still holding ``publicsuffixlist`` and ``qrcodegen`` back?

kfkitsune (approved):
Just curious about why we're holding back ``publicsuffixlist`` and ``qrcodegen``, but otherwise LGTM!

.github/dependabot.yml
charmander: They weren’t added because of Python 2 incompatibility.

- I linked qrcodegen’s commit log at the time, but I don’t remember why :D It might have been that most of the changes being made weren’t useful to us.
- publicsuffixlist upgrades make a lot of Dependabot noise since the public suffix list is always changing, and we don’t need the changes unless they affect domains we want to block.

PR #974, opened 2021-03-21
Post Python 3 Cleanup: Remove __future__ imports

kfkitsune:
The future is now, so let's remove ``__future__`` imports! 🎉

(Note: I think I cut this branch from ``upstream/master`` prior to the forced update removing c005258700509ce7444863a8cc5d82a8e4e3c4e5 (which would have set 8.1.2 as the version in use) -- I am not sure how to remedy this and remove that commit from this branch, but I did synchronize the ``Pillow`` versions with ``6.2.2``, as is currently on ``master``)


charmander (approved):
:tada:

PR #975, opened 2021-03-21
Docker fixes

kfkitsune:
Looking back to #885, a few lingering issues that are resolved (and eliminates the troubleshooting section entirely, and avoids re-symlinking ``sh`` to ``bash`` (which may or may not cause further issues, since Ubuntu might not expect it)):

- Adds ``storage/static`` as a created directory. (The nginx step would otherwise complain about this not existing.)
- Uncomments lines in ``site.config.txt.example``, permitting the template file to be used as-is without manually uncommenting the lines.
- Since Ubuntu is symlinking `/bin/sh` to `dash`, when on Ubuntu as a host invocations of the Docker scripts fails (`containers/run: 3: UID: parameter not set`). Converts usages of `$UID` to `$(id -u)`.

This was created by running through `README-docker.md` top-to-bottom, to ensure everything works as expected.

### DIFF
--- a/README-docker.md
+++ b/README-docker.md
@@ -101,7 +101,7 @@ containers/run \
 ## Nginx
 
 ```shell
-mkdir -p weasyl-web
+mkdir -p weasyl-web storage/static
 docker cp weasyl-web:/weasyl/build weasyl-web/
 containers/run-nginx
 ```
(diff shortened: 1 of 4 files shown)


charmander (approved):
Nice!

PR #976, opened 2021-03-21
Remove measurement of time a request spends queued

charmander:
because since Twisted 16.3.0, requests never spend time queued. See the relevant deprecation notices at https://twistedmatrix.com/documents/21.2.0/api/twisted.web.http.Request.html (“in 16.3 this method was changed to become a no-op, as Request objects are now never queued”) and https://twistedmatrix.com/trac/ticket/8320.

This should be followed up with the removal of `(queued: $sent_http_x_queued_time_spent)` from the Nginx timing log configuration.

- [ ] update Nginx timing log configuration (@charmander)

### DIFF
--- a/weasyl/middleware.py
+++ b/weasyl/middleware.py
@@ -51,19 +51,13 @@ def db_timer_tween_factory(handler, registry):
     """
     def db_timer_tween(request):
         started_at = time.time()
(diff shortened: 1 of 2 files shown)


kfkitsune (approved):
LGTM!

PR #977, opened 2021-03-21
Time request handling parts with monotonic `clock.perf_counter`

charmander:
Python 3!

In addition to being monotonic, operations on its return values are more precise, because zero is closer in time – but that effect is a decimal place or two past what we even display.


kfkitsune (approved):
LGTM!

PR #978, opened 2021-03-21
Remove forced password resets

charmander:
Confirmed in production that no accounts have this option set. There’s no UI to set it; it has to be done by ops. I think reducing the code complexity and amount of barely-exercised code, and satisfying this need with regular password resets, is an improvement.


kfkitsune (approved):
LGTM!

PR #980, opened 2021-03-22
Fix visual submission reuploads not regenerating WebP thumbnail

charmander:
I thought there was something to clear the link before it was switched to Pillow? That might have been `image_representations`.


kfkitsune (approved):
Yep, this works. Reuploading a submission with a new file now immediately regenerates the thumbnail, as expected.

LGTM!

PR #1046, opened 2021-06-19
Use docker-compose for development

charmander:
I hope this makes getting set up less painful for everyone. Right now things are configured in a sort of minimal way, which means somewhat long rebuild times after every change, but it does support use of a remote Docker instance out of the box, so people can set up Docker in, for example, a local or remote VM.

Closes #60.


kfkitsune (approved):
Nearest I can ascertain, this works flawlessly. (Had a moment of confusion when my browser added ``https`` as the protocol, and it failed to work, but other than that, no readily visible snags in getting it working.)

🎉

charmander:
Great to hear! \o/ Thanks for reviewing it.

charmander:
Should I remove the Vagrant-based setup as part of this? I don’t think it works anymore.

kfkitsune:
> 
> 
> Should I remove the Vagrant-based setup as part of this? I don’t think it works anymore.

Might as well; doesn't really make sense to maintain two distinct development environment toolkits, unless I am missing some non-obvious use case.

PR #1051, opened 2021-06-29
Remove inter-media links

charmander:
They add a lot of complexity along with other types of overhead, and are generated lazily in confusing ways while providing only a small benefit (not regenerating covers when identical submission files are reuploaded).

This slightly breaks the public HTTP API’s backwards-compatibility, but anything making use of the removed properties is already broken.

Can be deployed safely without running the migration, making it reversible (except for the missing links on new media).


kfkitsune (commented):
Nearest I can tell, it looks okay, and it seems to function in the Docker test environment; quick test of uploading visual, as well as multimedia and changing cover artwork for the latter seems to work okay.

PR #1054, opened 2021-06-30
Remove Google CAPTCHA

charmander:
We’re already going through Cloudflare and can use its captcha challenge on the `/signup` path.

Removing this means a better signup experience, less code complexity, slightly less exposure to third parties for Weasyl users, and a slightly nicer Content-Security-Policy when that’s implemented.


kfkitsune (approved):
LGTM!

PR #1055, opened 2021-06-30
Remove Dogpile cache key mangler

charmander:
It’s a disaster waiting to happen, since it maps distinct inputs to identical outputs. Luckily, the only arbitrary Unicode text passed to a cache function before this was tags, which discard information in a compatible way.

Preprocess before cache lookups to avoid creating keys incompatible with Memcached and also improve hit rate.

Also remove the unused configurable Memcached prefix.


kfkitsune (approved):
LGTM!

PR #1059, opened 2021-07-01
Add SVG favicon

charmander:
for high-DPI displays.

Traced manually from the logo, so it looks slightly different. Smaller than the PNG, even before compression! (318 bytes after Brotli.)


kfkitsune (approved):
LGTM!

PR #1063, opened 2021-07-02
Perform JSON serialization at the pylibmc client level; remove gzip

charmander:
I don’t think the gzip can be triggered by anything realistic – maybe 1MB of blocked tags. (But we should increase the object size limit anyway if that happens.)

I also tried packing the constant `ct` and `v` metadata separately, as well as msgpack, but they varied from slightly slower to much slower.

Development deployment:

- `./wzl restart memcached`

Production deployment:

- [ ] switch to new instance of previous version that only uses mcd3
- [ ] clear mcd4
- [ ] deploy new instance of new version that only uses mcd4
- [ ] clear mcd3
- [ ] switch to new instance of new version that uses both mcd3 and mcd4

### DIFF
--- a/etc/requirements.txt
+++ b/etc/requirements.txt
@@ -16,7 +16,7 @@ publicsuffixlist==0.6.13    # For extracting the base domain for the email black
 alembic==1.5.8
 arrow==0.15.2
 bcrypt==3.2.0
-dogpile.cache==1.0.2
+dogpile.cache==1.1.3
 #lxml ...
 oauthlib==2.1.0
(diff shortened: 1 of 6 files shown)


kfkitsune (approved):
Well, the site loads in the Docker environment; considering memcached is used heavily, and browsing the main page and bouncing back and forth between other pages didn't raise any errors, I think it's safe to say this works.

Not quite sure on an exact way to test the exact changes, but there's no obvious _changes_ (which makes sense, considering this is operating on the caching layer); so this LGTM!

PR #1064, opened 2021-07-03
Fix rating restriction on folder navigation between own submissions in SFW mode

charmander:
User-reported.


kfkitsune (approved):
LGTM; submissions with a rating higher than configured for SFW mode do not appear in the folder navigation for a user's own submissions; enabling it hides it in the folder view under the submission, and turning SFW mode off makes the submission reappear.

PR #1066, opened 2021-07-04
Remove styled file inputs

charmander:
They have problems:

- appearing empty when form state is restored, e.g. by returning to the page with browser back/forward
- no focus state
- a strange hover/click area
- inaccurate form validation message positioning

and the default browser appearance isn’t bad enough to justify the complexity of the current implementation.

### DIFF
--- a/assets/js/scripts.js
+++ b/assets/js/scripts.js
@@ -245,67 +245,6 @@
             updateChecked();
         });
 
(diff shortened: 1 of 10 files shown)


kfkitsune (approved):
LGTM. No remaining instances of ``styled-input`` seen via ``git grep styled-input``.

PR #1067, opened 2021-07-04
Assorted comment improvements

charmander:
- Label comment deletion action with “Delete” instead of “Hide”
- Allow deletion of comment with replies that are all deleted
- Fix new shout replies showing up as “undefined”

### DIFF
--- a/assets/js/scripts.js
+++ b/assets/js/scripts.js
@@ -1028,7 +1028,7 @@
                 children = null;
             }
 
-            if (confirm('Hide this comment and any replies?')) {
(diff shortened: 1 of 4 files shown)


kfkitsune (approved):
Visual review only; LGTM.

(Though I am not entirely certain if calling it 'deleted' is correct, either; deletion would, for the most commonly understood meaning, mean it's gone. And _technically_ it is gone in this implementation, though it's recoverable, instead of the content being deleted from the DB. Discussion for later, I guess.)

PR #1068, opened 2021-07-06
Remove fancy tag input

charmander:
It looks nice, at the cost of:

- ability to copy
- ability to edit tags after they’re created
- browser undo and redo
- probably more

Clients already provide whole-word selection/navigation/deletion, so nobody’s missing out on much by not having the “x” buttons anymore.

Also, 14px text instead of 13⅓px, because it felt right. I suspect that every field should be at least this size.


weasyl/templates/submit/character.html
```diff
@@ -57,7 +57,7 @@ <h3>Description</h3>
 
     <div class="column tagging">
         <h3>Tags</h3>
-        <textarea name="tags" class="input tags-textarea" rows="3"></textarea>
+        <textarea name="tags" class="input tags-textarea" rows="3" required></textarea>
```
kfkitsune, line 60: See comment on ``required`` for the modification to the journal code; I don't think we actually enforce this anywhere.

weasyl/templates/submit/multimedia.html
```diff
@@ -72,7 +72,7 @@ <h3>Description</h3>
 
     <div class="column tagging">
       <h3>Tags</h3>
-      <textarea name="tags" class="input tags-textarea" rows="3"></textarea>
+      <textarea name="tags" class="input tags-textarea" rows="3" required></textarea>
```
kfkitsune, line 75: Likewise, no Python enforcement here. See other comments.

kfkitsune (commented):
A few observations, but functionally it looks okay. Unless I am missing something, it does change the behavior for requiring tags for journal/character/multimedia submissions, though.

weasyl/templates/submit/journal.html
```diff
@@ -27,7 +27,7 @@ <h3>Content</h3>
 
     <div class="column tagging">
       <h3>Tags</h3>
-      <textarea name="tags" class="input tags-textarea" rows="3"></textarea>
+      <textarea name="tags" class="input tags-textarea" rows="3" required></textarea>
```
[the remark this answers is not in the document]
charmander: Eep, this one was accidental. Thanks!

weasyl/templates/submit/multimedia.html
charmander: I think we enforce this? Or the code seems to be trying to, at least. https://github.com/Weasyl/weasyl/blob/25699ec20c7fc62a8433524733ba2cd2d20e4674/weasyl/submission.py#L97-L98

charmander: Yep, checked; I get the error for new multimedia submissions on master.

kfkitsune (approved):
LGTM. It functioned okay prior to the change removing the ``required`` attribute from the journal/character tag fields to match current functionality, so this specific review is a visual review only.

(To be perfectly honest, my eyes kinda slid over the ``@_create_submission(expected_type=3)`` invocation for multimedia, which itself did have the tag quantity check. My bad, there.)

PR #1072, opened 2021-07-08
Upgrade Sentry client from raven to sentry-sdk

charmander:
The API isn’t great and there’s already a workaround for a bug/glaring omission, but the library is actively maintained and the new performance/tracing features are nice.

### DIFF
--- a/weasyl/wsgi.py
+++ b/weasyl/wsgi.py
@@ -1,6 +1,8 @@
+import sentry_sdk
 from pyramid.config import Configurator
 from pyramid.httpexceptions import HTTPNotFound
 from pyramid.response import Response
+from sentry_sdk.integrations.pyramid import PyramidIntegration
 
 from libweasyl.configuration import configure_libweasyl
 from weasyl.controllers.routes import setup_routes_and_views
@@ -40,7 +42,6 @@ setup_routes_and_views(config)
 # Setup properties and methods for request objects.
 config.add_request_method(mw.pg_connection_request_property, name='pg_connection', reify=True)
 config.add_request_method(mw.userid_request_property, name='userid', reify=True)
-config.add_request_method(mw.log_exc_request_method, name='log_exc')
 config.add_request_method(mw.web_input_request_method, name='web_input')
 config.add_request_method(mw.set_cookie_on_response)
 config.add_request_method(mw.delete_cookie_on_response)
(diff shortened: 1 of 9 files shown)


charmander:
@kfkitsune Just looking for a review to make sure the code looks sensible; probably no need for you to go through the inconvenience of setting up a local Sentry instance too. :)

weasyl/wsgi.py
```diff
@@ -54,7 +55,14 @@ def weasyl_404(request):
     wsgi_app = ProfilerMiddleware(
         wsgi_app, profile_dir=m.MACRO_STORAGE_ROOT + 'profile-stats')
 if config_obj.has_option('sentry', 'dsn'):
-    wsgi_app = mw.SentryEnvironmentMiddleware(wsgi_app, config_obj.get('sentry', 'dsn'))
+    sentry_sdk.init(
+        dsn=config_obj.get('sentry', 'dsn'),
+        release=d.CURRENT_SHA,
+        traces_sample_rate=float(config_obj.get('sentry', 'traces_sample_rate')),
+        integrations=[PyramidIntegration()],
+        send_default_pii=False,  # can’t be enabled as long as `before_send` doesn’t run for performance tracing!
```
kfkitsune, line 63: I looked over this earlier, and the comment is still breaking my brain (might just not be grokking something); the ``mw.strip_session_cookie`` function precludes this being ``True``, for performance reasons?

According to the [docs](https://develop.sentry.dev/sdk/data-handling/), one of the things this would permit is sending the user as Pyramid knows it; though we seem to override that by setting it ``set_user()`` on L99 in the middleware (this PR). It also seems to strip all cookies when ``False``, even though we sanitize ``WZL`` over in ``strip_session_cookie``.

kfkitsune (approved):
Since it's already in production, the code--overall--looks sane without setting up a standalone Sentry instance (technically Docker's an option).

One comment is breaking my brain, but on the whole it looks okay.

Also, ``todo/pyramid.md:L88`` could be set as done, since we're excising ``log_exc`` with this PR. ('Clean up use of log_exc(). [...]')

weasyl/wsgi.py
charmander: Not for performance reasons, for security reasons. `before_send` (i.e. `mw.strip_session_cookie`) runs for captured exceptions and messages, but *not* for performance tracing (enabled by the `traces_sample_rate` above). So with `send_default_pii=True`, performance tracing would have unredacted `WZL` cookies attached to it. The explicit `send_default_pii=False` is a workaround until there’s some way of filtering those events too.

PR #1073, opened 2021-07-08
Fix exchange rate fetching for Python 3

charmander:
Unnoticed until now because it’s disabled in both production and dev, but I tested it as part of the sentry-sdk migration. :)


kfkitsune (approved):
LGTM, considering ``requests.get(url).content`` would be a binary string, per ``requests``'s [documentation](https://docs.python-requests.org/en/master/user/quickstart/#binary-response-content). (Visual review only.)

PR #1074, opened 2021-07-08
Fix tall banners breaking layout

charmander:


charmander:
Actually, the bug might be that the banner isn’t resized by the backend. Need to check what’s promised by the settings text and the code.

charmander:
Also… did insufficiently wide banners used to be stretched? Because that seems like the right behaviour, and they aren’t now.

PR #1075, opened 2021-07-08
Index sorted inbox/outbox views

charmander:
`ORDER BY noteid` in `weasyl.note.select_inbox`/`select_outbox` without an index that includes `noteid` means PostgreSQL decides not to use the `userid`/`otherid` indexes, resulting in query times >1s to view some small inboxes.

Index `(otherid, noteid)` so `WHERE otherid = $1 ORDER BY noteid [DESC]` can use it; same for `userid`.


kfkitsune (approved):
Looks okay as far as I can tell!

PR #1079, opened 2021-07-09
Replace Twisted with Gunicorn

charmander:
It simplifies the code, and having a more common setup is nice.


kfkitsune (approved):
LGTM; seems to work as expected. (Given the nature of the twisted to gunicorn conversion, it's not really something that can be 'seen'.)

Pages load and submitting a test submission works correctly.

PR #1084, opened 2021-07-14
Login experience improvements

### LINKED ISSUE #697: Login shouldn’t redirect back to external pages
I haven’t confirmed this bug, but it looks like it exists:

https://github.com/Weasyl/weasyl/blob/29eb5d4c558403ffb46b4ae24217c67155ed1419/weasyl/controllers/user.py#L28

https://github.com/Weasyl/weasyl/blob/29eb5d4c558403ffb46b4ae24217c67155ed1419/weasyl/controllers/user.py#L47

Facilitates phishing.

charmander:
- Fixes search query (for example) disappearing when toggling SFW mode
- Fixes #697
- Removes potential point of poorly-handled failure: login modal XHR (also saving the corresponding delay)
- Makes logout a POST, and idempotent
- Removes unused frontend things

Known issue: the 2FA form’s redirect can still be specified with a URL parameter, and it still strips the querystring. Will be fixed in the upcoming 2FA refactor.

### DIFF
--- a/assets/js/scripts.js
+++ b/assets/js/scripts.js
@@ -172,9 +172,9 @@
 
             $('body').addClass('modal-login');
 
-            $('#login-box-container').empty().append(
+            $('#login-top').empty().append(
                 $('<div>', { id: 'login-box', class: 'content' }).append(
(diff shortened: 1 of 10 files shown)


kfkitsune (approved):
LGTM!

PR #1089, opened 2021-07-17
Rely on `Origin` header for CSRF prevention

charmander:
It’s supported by all modern browsers. Allows guests to be cookie-free and to share cache.

Some of the 2FA edge cases are probably broken (and it looks like they already were before), but I’d like to refactor that soon anyway.

SQLAlchemy is horrible as usual.

Intentionally missing a migration to delete `csrf_token` for now, just in case we want to undo this.

### DIFF
--- a/libweasyl/libweasyl/models/tables.py
+++ b/libweasyl/libweasyl/models/tables.py
@@ -633,7 +633,6 @@ sessions = Table(
     Column('sessionid', String(length=64), primary_key=True, nullable=False),
     Column('created_at', ArrowColumn(), nullable=False, server_default=text('now()')),
     Column('userid', Integer()),
-    Column('csrf_token', String(length=64)),
     Column('additional_data', JSONValuesColumn(), nullable=False, server_default=text(u"''::hstore")),
     Column('ip_address', String(length=39), nullable=True),
     Column('user_agent_id', Integer(), nullable=True),
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -7,7 +7,7 @@ jobs:
     runs-on: ubuntu-latest
 
     container:
-      image: ghcr.io/weasyl/ci-base-image@sha256:893c70251e56e2d443a132602c6f22d1d58fdf3f14fb77443d0d0d71a959e3b3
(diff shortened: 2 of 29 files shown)


libweasyl/libweasyl/models/tables.py
```diff
@@ -633,7 +633,6 @@ def default_fkey(*args, **kwargs):
     Column('sessionid', String(length=64), primary_key=True, nullable=False),
     Column('created_at', ArrowColumn(), nullable=False, server_default=text('now()')),
     Column('userid', Integer()),
-    Column('csrf_token', String(length=64)),
```
kfkitsune, line 636: We may be missing the alembic migration to remove the `csrf_token` column here.

```
weasyl=# select * from sessions;
       sessionid        |         created_at         | userid | csrf_token | additional_data |    ip_address     | user_agent_id 
------------------------+----------------------------+--------+------------+-----------------+-------------------+---------------
 SOr3kDbrUyiKp-L2KbpuqA | 2021-07-17 23:23:26.675837 |  26707 |            |                 | ::ffff:172.18.0.1 |             1
(1 row)

weasyl=# \q
could not save history to file "/root/.psql_history": No such file or directory
root@d77f21a0be0b:/# exit
kyra@kitty:~/git/weasyl$ git log
commit f0c0c62b412a1c4ba4f1d4b43fd62c546ab1f8a6 (HEAD, charmander/origin-csrf)
Author: Charmander <~@charmander.me>
Date:   Tue Jul 13 00:23:39 2021 -0700
```

kfkitsune (changes requested):
Don't see any reason why this wouldn't work, though it looks like the Alembic migration for the `csrf_token` didn't get added to the files committed; still seeing the column in the DB after a full rebuild of the containers.

libweasyl/libweasyl/models/tables.py
charmander: > Intentionally missing a migration to delete `csrf_token` for now, just in case we want to undo this.

I promise to put it in right after :D

kfkitsune: Missed that. My bad!

kfkitsune (approved):
LGTM, acknowledging that the migration to drop `csrf_token` is not yet included.

charmander:
Force-pushed removal of a leftover TODO comment that no longer applied.

charmander:
Thanks for reviewing this @kfkitsune! :)

PR #1092, opened 2021-07-20
Improve performance of submission, gallery pages

charmander:
for users with large numbers of submissions. ~300 ms reduced to ~10 ms for ~5,000 submissions, for example.

### DIFF
--- a/weasyl/submission.py
+++ b/weasyl/submission.py
@@ -901,44 +901,55 @@ def select_featured(userid, otherid, rating):
 
 def select_near(userid, rating, limit, otherid, folderid, submitid):
     statement = ["""
-        SELECT su.submitid, su.title, su.rating, su.unixtime, su.userid,
-               pr.username, su.subtype
+        SELECT su.submitid, su.title, su.rating, su.unixtime, su.subtype
           FROM submission su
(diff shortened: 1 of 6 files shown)


weasyl/submission.py
```diff
@@ -901,44 +901,55 @@ def select_featured(userid, otherid, rating):
...
     else:
-        statement.append(" AND su.rating <= %i AND su.settings !~ 'f'" % (rating,))
+        statement.append(" AND su.rating <= %(rating)s AND su.settings !~ 'f'")
 
     if folderid:
         statement.append(" AND su.folderid = %i" % folderid)
 
+    statement = "".join(statement)
+    statement = (
+        f"SELECT * FROM ({statement} AND su.submitid < %(submitid)s ORDER BY su.submitid DESC LIMIT 1) AS older"
```
kfkitsune, line 930: `f""`? That's new (had to look it up: PEP 498). So we're injecting the previously generated statement into this new one, if I am grokking this correctly?

kfkitsune (approved):
Looks okay to me! (Did learn about `f""`-strings, though!)

weasyl/submission.py
charmander: Yep!

PR #1095, opened 2021-07-23
Remove outdated index condition

charmander:
`favorites` has a `NOT NULL` constraint now.

Migration should run before app deployment.


kfkitsune (approved):
LGTM!

PR #1096, opened 2021-07-23
Add submission tag history index

charmander:
Oops.

Makes the submission tag history page not scan every row.


kfkitsune (approved):
LGTM!

Though I do wonder why `table_name` is a named parameter in `downgrade()`, but not `upgrade()`; doesn't affect the code, though.

PR #1097, opened 2021-07-26
Prepare to split journal and character hidden/friends-only into columns

charmander:
by initializing the columns to `NULL` for existing content/content created before the deployment of this app version, then filling them in and keeping them synchronized for future new, edited, and removed content.


kfkitsune (approved):
Issues in the docker environment aside--check Gitter for context--it looks okay to me.

PR #1098, opened 2021-07-26
Fix rendered Markdown trailing `<` caused by new `lxml.html.tostring` trailing newline

charmander:
Possibly introduced by a new libxml2 version?

Bug found by @kfkitsune.


kfkitsune (approved):
🎉 Yep, fixed! Merging!

PR #1103, opened 2021-08-04
Changes timezone designation for user sessions in admin view from PDT to UTC

kfkitsune:
Not quite sure when--if?--this changed, but the Docker environment now has logins showing as UTC for these session timestamps. So we should show the correct timezone, therefore. Found by @charmander.

(As far as I can tell, no code changes need to be made here; the time's showing up as UTC as-is.)


charmander:
Fixed by converting to the viewer’s local time in #1221.

PR #1107, opened 2021-08-16
Make Google Docs embedding stricter

charmander:
Already deployed, of course.

- prevents script injection
- stops users from accidentally embedding the wrong type of page for new submissions
- updates and simplifies the Google Docs embedding help page
- discourages embedding PDFs from Google Drive over uploading them directly


weasyl/templates/help/gdocs.html
```diff
@@ -1,50 +1,45 @@
...
-      embed code.
-    </p>
-
-    <figure class="help-image">
-      <img src="${resource_path('img/help/help-gdocs-4.png')}" width="540" height="88">
-    </figure>
+    <ol>
+      <li>
+        <figure class="help-image">
+          <figcaption>In the <kbd><samp>File</samp></kbd> menu of your document, there is a <kbd><samp>Publish to the web</samp></kbd> option.</figcaption>
```
kfkitsune, line 10: Question: Why ``<kbd><samp>...</samp></kbd>``? From a visual standpoint--currently, at least--it _feels_ like a more verbose ``<em>...</em>``/bolding; according to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd), the `<kbd>` element ``represents a span of inline text denoting textual user input from a keyboard, voice input, or any other text entry device. By convention, the user agent defaults to rendering the contents of a <kbd> element using its default monospace font, although this is not mandated by the HTML standard.``

Looking into ``<samp>`` as well, it ``is used to enclose inline text which represents sample (or quoted) output from a computer program``.

And I do see that the ``help.scss`` does set style on `kbd`/`samp`, but does this completely override User-Agent applied styles? (I'm mostly just theorizing at this point... from a usability perspective, it _shouldn't_ affect it, but it's something I noticed, was curious about, and thus I just figured I'd make a note.)

kfkitsune (approved):
Overall, LGTM. I am curious about using `kbd`/`samp` tags; I noted that in a distinct review comment (might just be missing some context for why they were selected as they were).

But overall, as-is, it looks good to merge!

weasyl/templates/help/gdocs.html
charmander: Lower on the page: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd#representing_onscreen_input_options

It’s a bit weird, but it’s specifically in [the HTML spec](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-kbd-element) too:

> When the `kbd` element contains a `samp` element, it represents input based on system output, for example invoking a menu item.

Supposedly just `kbd` is fine for the same purpose, but I figured I may as well. :)

PR #1112, opened 2021-08-23
Keep track of session last active timestamp

charmander:
Preparation for expiring old sessions.


kfkitsune (approved):
LGTM; functions as expected. Sessions without a `last_active` set, get set upon visiting a page (e.g., after deploying the code). Visiting after the threshold also updates the value stored.

PR #1124, opened 2021-09-07
Give more likely advice in CSRF check failure message

### LINKED ISSUE #1113: Error message for CSRF check failure should be improved
When users see this message, it’s usually because of an incompatible browser, not an actual CSRF attack. Give examples of compatible browsers.

charmander:
Most people who run into this do so because their browser doesn’t support the `Origin` header, not because they’re the victim of a CSRF attempt.

Fixes #1113.

### DIFF
--- a/weasyl/errorcode.py
+++ b/weasyl/errorcode.py
@@ -8,10 +8,7 @@ unsigned = "You must be signed in to perform this operation."
 permission = "You do not have permission to access this page."
 
 token = (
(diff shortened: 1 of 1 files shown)


kfkitsune (approved):
LGTM. Only real question is was the smart apostrophe intended? (``’`` vs. ``'``)

charmander:
It was intended, yep! We don’t use them consistently in site text yet, but we’ll get there. :)

PR #1212, opened 2022-04-12
Add CodeQL analysis

charmander:
Replacing LGTM.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,39 @@
+name: 'CodeQL'
+
+on:
+  push:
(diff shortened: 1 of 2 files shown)


kfkitsune (approved):
Might as well, if the core functions behind LGTM are default in GitHub now. (And it looks like the Semmle/LGTM team got folded into GitHub; so... another point for going native.)

At the risk of running into a meta-pun, LGTM.

As a query, though, does this add a status check entry in the Checks section (similar to how Codecov, LGTM, and the test actions do)? I'm not entirely certain how or when GitHub decides to integrate them, since it looks like the CodeQL ran for this specific PR.

PR #1221, opened 2022-05-06
Format timestamps in local time on client

charmander:
See commits for details! Mostly the last one.


kfkitsune (commented):
Visual review (might be a bit before I can get an in-VM review) looks okay. I'd certainly concur with letting the client render timestamps/zones based on the locale it is rendered in. (Less logic to handle on the backend, less data to store, fewer code paths.)

PR #1372, opened 2023-07-09
Improve tag suggestions

charmander:
- stops showing "Modify" to people who aren't logged in
- removes public tagging history and notifications to encourage suggesting tags
- removes the option to suggest tag removal
- adds approve/reject with feedback options for suggested tags

Undo tokens might be a temporary measure while all content types gain history tables, or they might become permanent, in which case improved secret management (separate storage, key rotation with `MultiFernet`) can be implemented later.

A list of tags that have been rejected for being abusive isn’t exposed anywhere yet, but I’ll check for them manually while the project to redo the entire mod UI is in progress.

Migration should run before web servers are switched over.

### DIFF
--- a/weasyl/searchtag.py
+++ b/weasyl/searchtag.py
@@ -582,11 +582,14 @@ def suggestion_arbitrate(
             case _:
                 raise TypeError()
 
-    undo_token = _undo_token_create(
-        target=target,
-        tagid=tagid,
-        added_by=check.added_by,
-    )
+    if check.added_by is None:
+        undo_token = None
+    else:
+        undo_token = _undo_token_create(
+            target=target,
+            tagid=tagid,
+            added_by=check.added_by,
+        )
 
     match action:
         case SuggestionAction.APPROVE:
--- a/weasyl/searchtag.py
+++ b/weasyl/searchtag.py
@@ -1,6 +1,13 @@
+import base64
+import enum
 import re
+import secrets
+import time
+from dataclasses import dataclass
+from struct import Struct
 
 import sqlalchemy as sa
+from cryptography.hazmat.primitives.ciphers.aead import ChaCha20Poly1305
 from sqlalchemy.sql.expression import any_
 
 from libweasyl import staff
@@ -10,7 +17,7 @@ from weasyl import define as d
 from weasyl import files
 from weasyl import ignoreuser
 from weasyl import macro as m
-from weasyl import welcome
+from weasyl.config import config_obj
 from weasyl.error import WeasylError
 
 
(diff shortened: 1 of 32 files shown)


weasyl/searchtag.py
```diff
@@ -176,152 +247,457 @@ def is_tag_restriction_pattern_valid(text):
...
+
+    def transaction(tx):
+        tx.execute(
+            "WITH"
+            f" removed AS (DELETE FROM {target.table}"
+            " WHERE targetid = %(target)s"
+            " AND settings !~ 'a'"
+            " AND added_by = %(user)s"
+            " AND tagid != ALL (%(tags)s) RETURNING tagid),"
+            f" added AS (INSERT INTO {target.table} (targetid, tagid, added_by)"
```
kfkitsune, line 304: I _think_ the issue is around here, but this looks to allow users to suggest a tag, and given that the `TaggableTarget.table` is updated (so `searchmapsubmit`, for example), this means that a tag that has been suggested--but not approved or rejected--can be searched for. (Unsure if this is intended behavior, or not.)

Consider the following sequence of actions:
- User A has a submission that they created, SubA.
- User B suggests the tag `FooBarBazQuz` to apply to SubA.
- Searching the website via the searchbar for `FooBarBazQuz` will return SubA as a result, despite not being approved or rejected.

For example, consider the following image; we're searching for the tag `qiakwidhchwiodfhdcf` which was suggested as a tag, yet viewing the submission itself shows that the tag does not exist.
image

kfkitsune (commented):
Two items that I _think_ might be issues; I don't know if the behavior is intended or not:
- There does not appear to be a mechanism to allow users to know tags have been suggested on their submitted items, due to the removal of `welcome.tag_update_insert()`.
- Suggested tags can be searched for without the owner of the item setting the tag as Approved.

Both items above have in-line comments.

charmander:
Thanks for checking it out, @kfkitsune! Both behaviors are intended. I thought it would strike a good usefulness balance. The differences between approved and suggested tags are that the latter show up at the end, in a different style, and not to guests. (The reason welcome type 3140 is still handled in the notification list is because this doesn’t delete existing notifications.)

kfkitsune:
Might be worth a distinct issue/PR to address, but I had to also edit `./wzl` to replace all instances of `docker-compose` with `docker compose` (dropping the hyphen); following the official Docker installation instructions for a fresh Debian 12 install does not provide the `docker-compose` command:
```shell
kyra@kitty:~/git/weasyl$ ./wzl configure
+ docker-compose run --rm -T configure alembic.ini
./wzl: 40: docker-compose: not found
kyra@kitty:~/git/weasyl$ sudo apt-get install docker-compose-plugin
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
docker-compose-plugin is already the newest version (2.19.1-1~debian.12~bookworm).
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
kyra@kitty:~/git/weasyl$ docker compose version
Docker Compose version v2.19.1
```

kfkitsune (approved):
Since the noticed items from the review are intended, LGTM!

charmander:
I’ve had to do that locally too for a while but didn’t commit it because of Debian backwards-compatibility! :D Okay, will commit a version that checks whether `docker-compose` exists.

PR #1386, opened 2023-12-04
Fix removal of blocked account e-mail domains for Python 3

charmander:
I refuse to write a regression test for this


charmander:
> All modified and coverable lines are covered by tests ✅

… no, they aren’t? This seems to be a Codecov bug caused by changing a line other than the first in a multi-line statement.

PR #1390, opened 2024-01-19
Add post counts to user tabs

charmander:
The counts respect the viewer’s rating preference. Sets up for some future changes:

* determining whether SFW mode has an effect when the configured rating preference is the same between SFW mode and non, so we can avoid confusingly showing a toggle that does nothing

* making blocked tag filtering a visible operation on pages of results (which itself sets up for better pagination and performance improvements through caching and lightened database load)

I’d like to provide counts for favorites and individual folders too, but they’re more involved. Ideally, counts would be stored in the database and updated eagerly; PostgreSQL makes it hard.

### DIFF
--- a/assets/scss/components/_user-tabs.scss
+++ b/assets/scss/components/_user-tabs.scss
@@ -1,0 +1,85 @@
+#user-nav > li {
+    width: 50%;
+    float: left;
+}
+
+#user-nav > li > a {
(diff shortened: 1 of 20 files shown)


kfkitsune (approved):
LGTM!

charmander:
Thanks for the review!

PR #1394, opened 2024-02-22
Update GitHub Actions actions to Node 20

charmander:
Node 16 actions are deprecated.

The Codecov action upgrade introduces a new upload token requirement, which pull requests from forks (which don’t receive the secret) are exempt from according to the action’s documentation.


charmander:
It looks like Codecov’s v4 action doesn’t support Alpine(?) yet. Installing a few packages (glibc and gnupg) in the base image might fix it, but this isn’t a priority and I don’t know if we even want to keep using Codecov.

charmander:
(even with gcompat,)

> Error relocating /tmp/_MEINbGjbD/libz.so.1: crc32_z: symbol not found

i’m so glad codecov upgraded to a “universal binary” :))

charmander:
CI logs:

> No token specified or token is empty

Yes, because the source is a fork. Uploads from forks aren’t supposed to require tokens, according to Codecov, for this very reason.

> There was an error running the uploader: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 429 - {'detail': ErrorDetail(string='Rate limit reached. Please upload with the Codecov repository upload token to resolve issue. Expected time to availability: 1642s.', code='throttled')}

What if I want to open more than one pull request every 30 minutes?

I think we’re dropping Codecov; I’ll just write an action that adds the coverage summary that we already have, plus maybe one from [diff-cover](https://pypi.org/project/diff-cover/), as a comment.

PR #1395, opened 2024-02-24
SFW mode improvements

### LINKED ISSUE #470: Site Preferences: "Maximum Viewable Content Rating in SFW Mode" should not include Explicit

charmander:
- Indicate when SFW mode does nothing and link to the page to configure it
- Remove option to set rating other than General for SFW mode

This should make the UI much clearer and easier to understand, at the expense of some niche configurability.

Fixes #470.

### DIFF
--- a/assets/scss/site.scss
+++ b/assets/scss/site.scss
@@ -1861,7 +1861,7 @@ $tag-reject-deemphasize-size: 11px;
     color: #999;
 }
 
-#header-utility span {
+#header-utility > span {
     display: inline-block;
     padding: 0 0.2em;
     color: #363f42;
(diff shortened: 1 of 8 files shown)


kfkitsune (approved):
LGTM, from a visual review.

charmander:
Thanks for the review @kfkitsune!

PR #1397, opened 2024-03-04
Make birthdate optional and less precise

charmander:
- Reduces friction when signing up
- Reduces amount of personal information stored by default
- Keeps exact birthdate private in case people don’t realize that displaying their age is equivalent to displaying their birthdate, as long as the transition is recorded

We still prevent users from displaying an age that’s inconsistent with anything they’ve done with ratings, and from changing their age once set.


libweasyl/alembic/versions/57171ee9e989_make_birthdate_optional.py
```diff
@@ -0,0 +1,34 @@
...
+    op.add_column('userinfo', sa.Column('asserted_adult', sa.Boolean(), server_default='f', nullable=False))
+    op.alter_column('userinfo', 'birthday',
+               existing_type=sa.INTEGER(),
+               nullable=True)
+
+
+def downgrade():
+    op.alter_column('userinfo', 'birthday',
+               existing_type=sa.INTEGER(),
+               nullable=False)
```
kfkitsune, line 30: Observation: Doesn't this cause an issue? If the upgrade step permits a null value in `userinfo`, wouldn't this constraint then fail for any users who have signed up after the upgrade is run in the event a downgrade is run after?

kfkitsune (commented):
Visually reviewed, and it mostly looks okay; however I do have a question about the Alembic migration... if we're altering if nulls are permitted in columns, doesn't that cause an issue if the downgrade needs to be run for any records where a null is present for the birthday?

(Unsure if this is actually an issue, but it's just something I observed when looking at the changeset.)

libweasyl/alembic/versions/57171ee9e989_make_birthdate_optional.py
charmander: Yes, it will fail. If we want to downgrade after people have signed up with this deployed, we’ll need to decide what we want to do with those accounts. But I hope we won’t want to downgrade :D

kfkitsune (approved):
LGTM!

charmander:
Thanks for the review, kfkitsune! I might try to add a few more tests before deploying this.

PR #1411, opened 2024-05-29
Don't copy root comment box's text to reply (#1106)

### LINKED ISSUE #1106: Comment Markdown preview breaks when opening Reply box with non-empty root comment input
Probably needs to be cleared after cloning.

huntertur:
Fixes #1106 by ensuring newly-created comment reply boxes do not contain what the user typed into the root comment box. This prevents `addMarkdownPreview` from trying to update the Markdown preview of a reply box that has not actually had its Markdown preview initialized.


charmander (approved):
Nice fix, thank you!

PR #1413, opened 2024-06-08
Add an explicit AI policy

charmander:
I tried to match the surrounding tone/formatting, even though I think it isn’t ideal. All the policies could probably use editing and tables of contents, but… later! Also resisted the urge to specify, wordily, that everything here is up to reasonable intent-based interpretation according to site staff – but like the rest of the rules, it is. (Yes, art is also “a process based on the work of people other than the author”.)

Hopefully this makes it clearer to everyone that, in short, “AI art” isn’t allowed on Weasyl.


huntertur:
I think the policy is reasonable!

I initially had a concern that "Any meaningful material generated by a process based on the work of people other than the creator of the overall piece" could inadvertently include collages people might make of art they've commissioned. In that case, even though the process is manual, it (cropping, copying, pasting) isn't adding any "new" information to the piece.

I can't think of any other case that could be slightly confusing at first glance.

charmander:
Thanks for the review, @huntertur! I think I’m happy to leave that situation up to the interpretation of the word “generated”, but if you or anyone have a good fix for it, then we can apply that.

ikanimew (approved):
This looks good!

PR #1423, opened 2024-08-22
Improve homepage’s display of recently popular submissions

charmander:
- #### Remove view count from recently-popular score calculation

    If two submissions have the same number of favorites, is the more viewed one more popular? If a submission accumulates views by virtue of being in Recently Popular without also accumulating favorites, is that a positive signal? I don’t think so.

    Views are a very flawed metric anyway, especially as currently implemented. (Also, with Weasyl’s current low traffic level, an advantage of 25 views can outweigh a 40% advantage in favorites near the top of Recently Popular.)

    Migration runs before deployment.

- #### Switch latest posts with recently popular on home page

    Recently Popular tends to be more appealing(!) and less risky viewing. Maybe in the future, we can exclude people the viewing user already follows, to make it still useful for discovering new artists.

    This is also good for cache hit rates.

### DIFF
--- a/assets/scss/site.scss
+++ b/assets/scss/site.scss
@@ -2256,7 +2256,7 @@ $notifs-color-active: #b9b9b9;
 
 /* ------------------------------------ =homepage --*/
 
-.home-latest {
+.home-top {
     margin: 8px 0;
 }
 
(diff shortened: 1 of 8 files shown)


huntertur:
* Looking at your proposal, I agree with you that dropping views makes sense. If I am understanding the view counting logic correctly (`weasyl.define.common_view_content`), IP addresses are used to make views unique for logged-out accounts. It's a lot easier to get a ton of IP addresses to hit a submission for popularity manipulation purposes than it is to create that many accounts, verify them, and just favorite a single submission as a bot. (Also it looks like there is an anti-botting measure implemented in `turnstile`? If so, that would almost completely seal the door against someone trying to manipulate the Recently Popular list.)
* I really like the idea to switch Recently Popular and latest posts so that the former is on top. I think that makes a lot of sense!

PR #1424, opened 2024-08-23
Separate metrics for retrieving notification counts from overall request metrics

charmander:
Uncached notification count queries can be triggered unpredictably by almost every page and take a highly variable amount of time depending on the user (up to 7 seconds). Keeping this separate should make individual route metrics more useful.


charmander:
I deployed this ahead of time, and it didn’t have as much of a clear impact on most of the graphs as I had hoped, but it did at least achieve the main goal of eliminating the peaks on the order of 7 seconds caused by notification counting from non-notification-related routes (mostly `index`).

Issue #1110, opened 2021-08-19
Rethink “Back” buttons


charmander:
e.g. on error messages and some forms – the ones that just do `history.back()`.

- Browsers already have built-in means of history navigation. We want people to use this.
- Their existence could make phishing easier, if a user makes the somewhat reasonable assumption that all form buttons on Weasyl lead to Weasyl. A quick redirect through a page that imitates Weasyl when navigated back to, and voilà.

     We should probably fix that even more thoroughly, though (because browser Back is also susceptible) by implementing forms of passwordless login. The ideal is for *everyone* to be using either password managers with generated passwords or going passwordless, and for our second factors to be unphishable.

     Removing special permissions from staff accounts and requiring them to use a separate interface also helps with this and more.

huntertur:
I see some pages, such as the art upload form, instead have a "Return Home" button: https://www.weasyl.com/submit/visual

My opinion is that removing all "Back" buttons would be best since they are redundant with the browser's back button, and the user can still navigate to other parts of the website through the header and footer. However, they could also be converted to Return Home buttons to match the precedent set by the upload pages. Do you have a strong opinion toward whether or not to keep these buttons?

charmander:
@huntertur I agree that removing them and letting people learn/rely on the global site navigation seems best.

PR #1432, opened 2024-10-06
Remove "Back"-esque buttons (#1110)

huntertur:
I know the original issue (#1110) only talked about removing buttons that acted like `history.back();`, but based on the discussion about preferring people to use the back button in their browser and global site navigation, I looked for and removed many of the "back"-style buttons. Please let me know if this is way too overzealous and I can trim this back.


charmander:
Nice improvement! The diff looks good, and I’m going to double-check that any now-standalone `float: right` buttons don’t change the layout to finish out the review. (There’s a lot of preexisting containment/`clear: both`, so it’s not hugely likely, but still!)

Issue #1104, opened 2021-08-05
Indicate whether content is for friends only on its thumbnail or equivalent


charmander:
… maybe?

PR #1433, opened 2024-10-06
Implement friends-only indicator (#1104)

huntertur:
Implements #1104.

Some notes:
- I added the friends-only indicator everywhere the existing mature/explicit indicator exists. The exception is in the critique section of the homepage, since I believe the underlying SQL query there explicitly filters out all friends-only submissions, even if the current user can see them?
- Since this is less important than the mature/explicit indicator, I made the new one about 2/3 the size of the existing one. This also solves a space constraint problem when displaying journals.
- I noticed that a user's journal page (`/journals/username`) doesn't currently show mature/explicit indicators, so I left that page alone. Maybe I could open an issue for that one and take a look into adding indicator support in general as a separate item.
- I chose `pink` instead of eyedropper-picking the existing color for Friends Only because I feel it is too close to the color used by the explicit indicator and could be therefore be confusing.

# Examples of how this looks

image

image

image

image

### DIFF
--- a/assets/scss/site.scss
+++ b/assets/scss/site.scss
@@ -1017,6 +1017,19 @@ a:hover .icon-report, button:hover .icon-report { background-position: -20px -24
     text-indent: 150%;
     z-index: 3;
 }
+.avatar-grid .friends-only {
+    width: 33px;
+    height: 33px;
+    border-style: solid;
+    border-width: 4px 4px 0 0;
+    position: absolute;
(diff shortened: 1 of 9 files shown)


philpem:
I noticed that the https://www.weasyl.com/journals/(username) doesn't show the rating or friends-only status of journals. If @charmander thinks it's needed, it'd be nice to get it in on this PR.

PR #1438, opened 2024-10-28
Start adding type annotations to libweasyl

huntertur:
I focused on adding type annotations to method parameters and return types. `STYLE_GUIDE.md` does not put any constraints on Python type annotation usage, so I went with the following approach:

- For methods that don't return anything, I left the return type unannotated instead of explicitly stating `-> None`.
- Parameters with defaults were left unannotated since the type is implied by the default value, unless the default value doesn't provide enough type information.
- Unused parameters were left unannotated.
- `__dunder__` methods were left unannotated since my understanding is they are well-known by everything (I could be wrong).
- `self` and `cls` were left unannotated since their types are implied.
- I didn't touch test code.
- I didn't touch `kwargs` since it looks like `typing.Unpack` requires Python 3.11, which is too new.

# Ignored portions

- Not touching methods that just do SQLAlchemy stuff or Pyramid stuff per discussion in the developer chatroom.
- Not touching `images`, `images_new`, or most of `models.media` per discussion in the developer chatroom about sanpera and PIL being replaced with a Rust-based imaging component.
- Not touching interface implementations (`cache`, `oauth`) since my understanding is tools will use the parent class for type information (I could be wrong), and my IDE seems to agree.

# Things I am unsure about

- Stuff related to `media` has been a pain. I tried to do my best, but it feels really messy, and I see that higher layers of code add additional key-value pairs onto `MediaItem.serialize`. I am open to feedback on this, even if the feedback is to hit the undo button on this and the associated changes in `models.media`. Part of this also extends to `configuration.configure_libweasyl`, specifically for `media_link_formatter_callback`.

# Closing thoughts

I started with `libweasyl` since it seems like a manageable piece of the project to bite off in one go. Unsure when I would start on `weasyl`, or how many sub-PRs I would break that into, but starting with `libweasyl` allows me to absorb feedback on how to best proceed with this type of change in the future.

### DIFF
--- a/libweasyl/cache.py
+++ b/libweasyl/cache.py
@@ -9,7 +9,6 @@ project.
 
 import json
 import threading
-from typing import Any
 
 import dogpile.cache
 import dogpile.cache.backends.memcached
@@ -51,7 +50,7 @@ class ThreadCacheProxy(ProxyBackend):
             pass
 
     @property
-    def _dict(self) -> dict[str, Any]:
+    def _dict(self):
         """
         Get the cache dict for the current thread.
 
@@ -62,7 +61,7 @@ class ThreadCacheProxy(ProxyBackend):
             self._local.cache_dict = {}
         return self._local.cache_dict
 
-    def get(self, key: str) -> Any:
+    def get(self, key):
         """
         Proxy a ``get`` call.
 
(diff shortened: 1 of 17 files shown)


huntertur:
Copy-pasting from developer chatroom:

Ohhhh okay I better understand why adding type annotations everywhere isn't wholly feasible. I was aware of mixins on the model side of things, but now I see the existence of things like `config.add_request_method` that would likely result in not being able to accurately type things

PR #1439, opened 2024-10-28
Add Bluesky as a social media site

huntertur:


charmander (approved):
This is included in part of some upcoming changes related to Bluesky + Mastodon + profile links, but getting it deployed sooner is a good idea.

PR #1443, opened 2024-10-30
Add links to Weasyl’s Bluesky and Mastodon profiles to footer

charmander:
including `rel="me"` on the homepage so the link back to Weasyl can be displayed as [verified](https://docs.joinmastodon.org/user/profile/#verification).

I originally wanted to list the actual handles like in the weasyl3 design (with a new offset underline also applied to the support e-mail address to avoid breaking the ink on `@`), but the Mastodon handle is too long and introduces text wrapping issues.


huntertur (approved):

huntertur:
Looks good :) `rel="me"` is a great idea!

charmander:
Thanks for the review!

philpem:
> Looks good :) `rel="me"` is a great idea!

It's needed for Mastodon url-verification :)

Issue #1447, opened 2024-10-31
Weasyl blocks users over the age of 100


philpem:
Spotted this while looking at the recent test fails due to issues with the 13th and 18th birthday tests...

It looks like we block users over the age of 100. The oldest recorded human was 122 years and 164 days: https://www.washingtonpost.com/world/2019/01/12/how-madame-calment-worlds-oldest-person-became-fuel-russian-conspiracy-theory/?noredirect=on

https://github.com/Weasyl/weasyl/blob/c3c4c280db5a461b506493d457dd5e58cdeeb483/weasyl/profile.py#L576

Maybe just set it to 125?

PR #1448, opened 2024-10-31
profile.py: allow ages over 100

philpem:
Oldest recorded human was 122 years 164 days but we're pretty sure she wasn't a Weasyl user. Either way, it might happen and we shouldn't be ageist.

(no, this comment isn't entirely serious but we might as well fix this)

related to #1447


charmander (changes requested):
This needs a corresponding `min` change to the field in `weasyl/templates/control/edit_profile.html`.

charmander (approved):

PR #1450, opened 2024-10-31
Fix error when searching for users with symbols (#416)

### LINKED ISSUE #416: User search: Cannot search for e-mail addresses containing "+"
Attempting to search by e-mail address for any string containing "+" throws an error.

Example request-id: 8a7061be725a464c9e6c3c3f3b8dfaa6-SF+Rg9h6

huntertur:
Fixes #416 

~~Double-backslash prevents the error case you brought up in the issue.~~

### DIFF
--- a/weasyl/moderation.py
+++ b/weasyl/moderation.py
@@ -364,10 +364,10 @@ def finduser(targetid, username, email, dateafter, datebefore, excludesuspended,
     if targetid:
         q = q.where(lo.c.userid == targetid)
     elif username:
-        q = q.where(lo.c.login_name.op('ilike')('%%%s%%' % username.replace('\\', r'\\')))
+        q = q.where(lo.c.login_name.ilike(f"%{username}%", escape=""))
     elif email:
         q = q.where(sa.or_(
-            lo.c.email.op('ilike')('%%%s%%' % email.replace('\\', r'\\')),
+            lo.c.email.ilike(f"%{email}%", escape=""),
         ))
 
     # Filter for banned and/or suspended accounts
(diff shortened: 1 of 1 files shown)


weasyl/moderation.py
```diff
@@ -364,11 +364,10 @@ def finduser(targetid, username, email, dateafter, datebefore, excludesuspended,
     if targetid:
         q = q.where(lo.c.userid == targetid)
     elif username:
-        q = q.where(lo.c.login_name.op('~')(username))
+        q = q.where(lo.c.login_name.op('ilike')('%%%s%%' % username.replace('\\', r'\\')))
```
charmander, line 367:
```suggestion
        q = q.where(lo.c.login_name.ilike(f"%{username}%", escape=""))
```

weasyl/moderation.py
```diff
@@ -364,11 +364,10 @@ def finduser(targetid, username, email, dateafter, datebefore, excludesuspended,
...
     elif email:
         q = q.where(sa.or_(
-            lo.c.email.op('~')(email),
-            lo.c.email.op('ilike')('%%%s%%' % email),
+            lo.c.email.op('ilike')('%%%s%%' % email.replace('\\', r'\\')),
```
charmander, line 370:
```suggestion
            lo.c.email.ilike(f"%{email}%", escape=""),
```

charmander (commented):
readability improvement?

weasyl/moderation.py
philpem, line 367: Are we on a recent enough python to use f-strings e.g. `f"%{username_r}%"` here, and break out the `username.replace` into a string assignment above?

charmander: Actually, this doesn’t work with SQLAlchemy >:(

huntertur: Yeah, f-strings are 3.6, and the Dockerfile is for 3.10

charmander (approved):

PR #1451, opened 2024-10-31
docker-compose: Quote environment strings in Docker Compose config

### LINKED ISSUE #1449: Docker compose script contains unquoted booleans
The docker-compose script fails on docker-compose 1.29.2 / Linux Mint 21.3, with the following error:

```
~/dev/weasyl$ ./wzl configure
+ docker-compose run --rm -T configure alembic.ini
ERROR: The Compose file './docker-compose.yml' is invalid because:
services.web.environment.WEASYL_RELOAD_ASSETS contains true, which is an invalid type, it should be a string, number, or a null
```

Looks like there are some unquoted strings in the `docker-compose.yml`'s `environment` blocks.

philpem:
Fixes #1449

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -49,13 +49,13 @@ services:
       args:
         version: '${WEASYL_VERSION:-}'
     environment:
-      WEASYL_STORAGE_ROOT: storage
-      WEASYL_RELOAD_ASSETS: true
-      WEASYL_RELOAD_TEMPLATES: true
-      WEB_CONCURRENCY: 8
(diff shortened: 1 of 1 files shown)


philpem:
I've also noted in the README that the Docker distribution packages don't have a working BuildKit on Ubuntu 24 LTS and derivatives. Bleh bleh bleh etc.

charmander:
(updated this pull request to do just what it says ’cause that’s best practice and i have a couple of nitpicks with the readme changes)

PR #1454, opened 2024-11-01
Fix 500 when visiting /note with no noteid (#248)

### LINKED ISSUE #248: AttributeError: 'noteid' when visiting ``/note`` route without form data

huntertur:
Fixes #248

I took the opportunity to practice writing tests for Weasyl by adding `weasyl.test.web.test_interaction`.

### DIFF
--- a/weasyl/controllers/interaction.py
+++ b/weasyl/controllers/interaction.py
@@ -4,7 +4,7 @@ from pyramid.response import Response
 from weasyl.controllers.decorators import login_required, token_checked
 from weasyl.error import WeasylError
 from weasyl import (
-    define, favorite, followuser, frienduser, ignoreuser, note, pagination, profile)
+    define, favorite, followuser, forms, frienduser, ignoreuser, note, pagination, profile)
 
 
 # User interactivity functions
@@ -97,9 +97,9 @@ def note_(request):
     if not define.is_vouched_for(request.userid):
         raise WeasylError("vouchRequired")
 
-    form = request.web_input()
+    noteid = forms.expect_id(request.GET.get("noteid"), "noteidInvalid")
 
-    data = note.select_view(request.userid, int(form.noteid))
+    data = note.select_view(request.userid, noteid)
 
     return Response(define.webpage(request.userid, "note/message_view.html", [
         # Private message
--- a/weasyl/controllers/interaction.py
+++ b/weasyl/controllers/interaction.py
@@ -97,7 +97,10 @@ def note_(request):
     if not define.is_vouched_for(request.userid):
         raise WeasylError("vouchRequired")
 
-    noteid = forms.expect_id(request.GET.get("noteid"), "noteidInvalid")
+    try:
+        noteid = forms.expect_id(request.GET.getone("noteid"))
+    except (KeyError, WeasylError) as e:
+        raise WeasylError("noteRecordMissing") from e
 
     data = note.select_view(request.userid, noteid)
 
--- a/weasyl/errorcode.py
+++ b/weasyl/errorcode.py
@@ -98,6 +98,7 @@ error_messages = {
     'noGuests': no_guest_access,
     "noCover": "No cover exists for that submission.",
     "noImageSource": "No image exists from which to create a thumbnail.",
+    "noteidInvalid": "The specified note is invalid.",
     "notEnoughTags": (
         "You must maintain at least two tags on this content in order to sufficiently describe it "
         "for our search system."),
@@ -176,6 +177,8 @@ error_status_code = {
     "submissionRecordMissing": 404,
     "journalRecordMissing": 404,
     "characterRecordMissing": 404,
+    "noteRecordMissing": 404,
+    "noteidInvalid": 403,
     "vouchRequired": 403,
     "signed": 403,
     "unsigned": 403,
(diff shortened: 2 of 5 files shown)


weasyl/errorcode.py
```diff
@@ -98,6 +98,7 @@
     'noGuests': no_guest_access,
     "noCover": "No cover exists for that submission.",
     "noImageSource": "No image exists from which to create a thumbnail.",
+    "noteidInvalid": "The specified note is invalid.",
```
charmander, line 101: I think a new error code for every invalid parameter is overkill. (It’s also not a 403.)

weasyl/forms.py
```diff
@@ -6,15 +6,16 @@
 _T = TypeVar("T")
 
 
-def expect_id(s: str) -> int:
+def expect_id(s: str, error_message="Unexpected") -> int:
```
charmander, line 9: No changes to this function, please. Keep it simple. `isinstance(s, str)` shouldn’t be checked on `s: str`.

weasyl/test/web/test_interaction.py
```diff
@@ -0,0 +1,72 @@
...
+from weasyl import errorcode
+from weasyl.test import db_utils
+
+
+INVALID_PATHS = (
+    '/note',
+    '/note?',
+    '/note?noteid',
+    '/note?noteid=',
+    '/note?noteid=a',
```
charmander, line 12: These tests are also overkill for something that’s more about integer parsing than notes.

charmander (changes requested):

weasyl/errorcode.py
huntertur: Would you prefer a switch to the existing `noteRecordMissing` ("That note doesn't seem to be in our database."), and if so, would you want this to be a 404 or remain the default 422?

weasyl/forms.py
huntertur: Will revert this.

weasyl/test/web/test_interaction.py
huntertur: OK if I move these to a separate `test_forms.py` that just tests `expect_id` directly, or would you want most of these gone entirely?

weasyl/test/web/test_interaction.py
```diff
@@ -0,0 +1,66 @@
...
+@pytest.mark.parametrize('path', INVALID_PATHS)
+def test_get_note_invalid(app, path):
+    user = db_utils.create_user()
+    cookie = db_utils.create_session(user)
+    resp = app.get(path, headers={'Cookie': cookie}, status=404)
+    assert resp.html.find(id='error_content').p.text.strip() == errorcode.error_messages['noteRecordMissing']
+
+
+@pytest.mark.usefixtures('db')
+def test_get_note_permissions(app):
```
charmander, line 30: Thanks for adding this test! This is a good test.

weasyl/test/web/test_interaction.py
```diff
@@ -0,0 +1,66 @@
...
+        'content': 'Content 1',
+    }, headers={'Cookie': cookie1})
+
+    app.post('/notes/compose', {
+        'recipient': 'user3',
+        'title': 'Title 2',
+        'content': 'Content 2',
+    }, headers={'Cookie': cookie2})
+
+    resp = app.get('/note?noteid=1', headers={'Cookie': cookie1})
```
charmander, line 50: Can’t rely on these notes having a particular id. As far as how to find that id… exercising `/notes` might not be a bad thing?

weasyl/test/web/test_interaction.py
```diff
@@ -0,0 +1,66 @@
+import pytest
+
+from weasyl import errorcode
+from weasyl.test import db_utils
+
+
+INVALID_PATHS = (
+    '/note',
+    '/note?noteid=1&noteid=2',
```
charmander, line 9: Again, not really a thing that’s specific to notes. Regression tests are good in general, but extrapolating from this one, we will be testing basic parameter handling on every route.

weasyl/controllers/interaction.py
```diff
@@ -97,9 +97,12 @@ def note_(request):
     if not define.is_vouched_for(request.userid):
         raise WeasylError("vouchRequired")
 
-    form = request.web_input()
+    try:
+        noteid = forms.expect_id(request.GET.getone("noteid"))
+    except (KeyError, WeasylError) as e:
```
charmander, line 102: Nice try – this is about is good as it can be.

charmander (changes requested):
I’d still recommend just not fixing or testing this particular endpoint standalone – there’s nothing about `/note` that merits this special treatment as far as I’m aware. But if you could separate out `test_forms` and `test_get_note_permissions` into their own PRs, those would be great.

weasyl/test/web/test_interaction.py
huntertur: This is fair. I will remove the `INVALID_PATHS`-related tests and keep `test_get_note_permissions`.

huntertur:
Closing this PR to split out the new tests into their own PRs.

PR #1457, opened 2024-11-01
Add Sphinx for project documentation

kfkitsune:
This can be tweaked later, but this should provide a decent 'starting point', for lack of a better term.

So assuming I have this correct, this should be automatically creating the documentation for both `libweasyl` and `weasyl`, and when pushed to the `main` branch, the added workflow will publish to [GitHub Pages](https://www.sphinx-doc.org/en/master/tutorial/deploying.html#id5).

Not quite sure where to add the command to rebuild the HTML; to be honest, the fact that we're abstracting packages into a Docker container is a tad cumbersome, and to get it built as it was, I needed to have a venv on my host via Poetry, and `poetry install --with=docs --sync` to get Sphinx installed to be able to build it. There's probably some little thing I am overlooking, but... I'm hoping this provides a good starting point.


charmander:
I appreciate the effort, and the rendered output looks very nice, but I don’t think the added machinery is worth it.

kfkitsune:
> The additional developments docs are good, though, and would be a good thing to link to as a normal Markdown document somewhere in the README, without duplication.

Do you mean the additional information in `docs/source/commands.md`?

Also it's not a problem; this is easy enough to recreate if at some point in the future we want Sphinx documentation. The Sphinx extensions I used for this did the vast majority of the heavy lifting, and the only real complexity was having to deal with with `wzl`, since for Sphinx to build the documentation `weasyl` and `libweasyl` installed for the documentation to build (which means installing the `docs` dependencies as they are needed).

PR #1467, opened 2024-11-04
Improve tag blocking UI (#1453)

### LINKED ISSUE #1453: Improve tag blocking UI
- separate forms for adding and removing, because Rating to Filter only applies to adding – either put a Remove button next to every entry, or use checkboxes (a `<table>` is probably a good idea either way)
- remove outdated advice and labelling about blacklist filters (we used to also have whitelist filters); just call the entire feature “blocked tags”
- figure out what should happen when someone tries to add a duplicate tag (currently nothing, regardless of associated ratings); should the control to update a blocked tag’s rating bound be part of the table (radio buttons, dropdowns)?
- remove cancel button
- allow adding multiple tags at once?

huntertur:
Closes #1453 

# Before
image

# After
image


charmander:
Thoughts on “Block general and higher” → “Block all ratings”?

huntertur:
I like "block all ratings"; it sounds less cumbersome than "block general and higher". 

My one concern is that, in the (admittedly unlikely) event that a rating lower than General is added to the website, tags which were thought to be blocked on all ratings would suddenly no longer be blocked on all tags, since they would start to appear on pre-General content.

I know the prospect of a pre-General rating getting added is very low, but since it sounds like the rating structure may have changed at least once in the past, I bring this up just in case. If you are OK with this, I will make the change, plus the form field name change suggested in the developer chatroom.

Additional concern while reviewing the text further: I notice that elsewhere on the website, ratings seem to usually be referred to with a capital letter; General and Mature and Explicit are often written as such, but they are lowercase here. If I make this change and your suggested change, the list will look like this:

* Block all ratings
* Block Mature and higher
* Block Explicit only

charmander:
> My one concern is that, in the (admittedly unlikely) event that a rating lower than General is added to the website, tags which were thought to be blocked on all ratings would suddenly no longer be blocked on all tags, since they would start to appear on pre-General content.

Yeah, I’m okay with that. On top of that situation being unlikely, I can’t imagine we would do anything in it other than migrate General+ tag filters to also block the new rating.

For what it’s worth, *my* main concern was that it would stop matching the column. I think keeping General+ in the column is fine, though.

> Additional concern while reviewing the text further: I notice that elsewhere on the website, ratings seem to usually be referred to with a capital letter; General and Mature and Explicit are often written as such, but they are lowercase here.
>
> If I make this change and your suggested change, the list will look like this:
>
> - Block all ratings
> - Block Mature and higher
> - Block Explicit only

This list looks good. (In general, the site overuses title case, and it’s not *necessary* here, but I have a slight preference for your suggested version.)

weasyl/searchtag.py
```diff
@@ -178,6 +178,27 @@ def get_or_create(name):
     return _get_or_create(d.get_search_tag(name))
 
 
+@region.cache_on_arguments()
```
charmander, line 181: This will use the entire list as the cache key – see existing uses of `cache_multi_on_arguments`, which can be folded into the above as

```python
def _get_or_create(*name: str) -> list[str]: ...
```

weasyl/searchtag.py
```diff
@@ -178,6 +178,27 @@ def get_or_create(name):
...
+        SELECT title
+        FROM UNNEST(%(names)s::text[]) AS title
+        ON CONFLICT (title) DO NOTHING
+    ''', names=names)
+
+    rows = d.engine.execute(
+        'SELECT tagid FROM searchtag WHERE title = ANY (%(names)s)',
+        names=names)
+
+    return list(row[0] for row in rows)
```
charmander, line 194:
```suggestion
    result = d.engine.execute(
        'SELECT tagid FROM searchtag WHERE title = ANY (%(names)s)',
        names=names)

    return result.scalars().all()
```

weasyl/controllers/settings.py
```diff
@@ -712,19 +712,20 @@ def manage_tagfilters_get_(request):
...
         profile.get_user_ratings(request.userid),
-    ], title="Tag Filters"))
+    ], title="Blocked Tags"))
 
 
 @login_required
 @token_checked
 def manage_tagfilters_post_(request):
     do = request.POST["do"]
-    title = request.POST["title"]
```
charmander, line 722: Form compatibility, discussed in chat

charmander (changes requested):

huntertur:
Pushed new commits to address code review feedback and retain form compatibility.

weasyl/searchtag.py
```diff
@@ -160,24 +160,33 @@ def select_list(map_table, targetids):
...
-        name=name)
+    return result.scalars().all()
 
 
 def get_or_create(name):
     return _get_or_create(d.get_search_tag(name))
 
 
+def get_or_create_many(names: list[str]) -> list[int]:
+    normalized_names = map(d.get_search_tag, names)
```
charmander, line 186: I know it’s a little unsatisfying because this is named similar to `get_or_create`, but given that this doesn’t deduplicate or remove empty tags and the value it’s passed is already normalized, how about this patch:

```patch
diff --git a/weasyl/blocktag.py b/weasyl/blocktag.py
index 11930804..b9dea48d 100644
--- a/weasyl/blocktag.py
+++ b/weasyl/blocktag.py
@@ -79,7 +79,7 @@ def insert(userid: int, tags: str, rating: int):
 
     profile.check_user_rating_allowed(userid, rating)
 
-    parsed_tags = list(searchtag.parse_tags(tags))
+    parsed_tags = searchtag.parse_tags(tags)
 
     d.engine.execute('''
         INSERT INTO blocktag (userid, tagid, rating)
diff --git a/weasyl/searchtag.py b/weasyl/searchtag.py
index 0275ef0d..71443a46 100644
--- a/weasyl/searchtag.py
+++ b/weasyl/searchtag.py
@@ -5,6 +5,7 @@ import secrets
 import time
 from dataclasses import dataclass
 from struct import Struct
+from typing import Iterable
 
 import sqlalchemy as sa
 from cryptography.hazmat.primitives.ciphers.aead import ChaCha20Poly1305
@@ -182,8 +183,7 @@ def get_or_create(name):
     return _get_or_create(d.get_search_tag(name))
 
 
-def get_or_create_many(names: list[str]) -> list[int]:
-    normalized_names = map(d.get_search_tag, names)
+def get_or_create_many(normalized_names: Iterable[str]) -> list[int]:
     return _get_or_create(*normalized_names)
 
 
```

charmander (approved):

weasyl/searchtag.py
huntertur: That's a good idea! Applied that now.

charmander (approved):

PR #1469, opened 2024-11-05
Platform upgrades

charmander:


imagemagick-policy.xml
```diff
@@ -0,0 +1,85 @@
...
+  <policy domain="resource" name="disk" value="0"/>
+  <!-- Set the maximum length of an image sequence.  When this limit is
+       exceeded, an exception is thrown. -->
+  <policy domain="resource" name="list-length" value="3600"/>
+  <!-- Set the maximum width of an image.  When this limit is exceeded, an
+       exception is thrown. -->
+  <policy domain="resource" name="width" value="16384"/>
+  <!-- Set the maximum height of an image.  When this limit is exceeded, an
+       exception is thrown. -->
+  <policy domain="resource" name="height" value="16384"/>
```
huntertur, line 53: Could the 16,384px width and height limits cause any issues for long comics that artists wish to upload as one piece?

I am aware of at least one example of a submission getting near these limits; 15,550px-height image here: https://www.weasyl.com/~trevorfox/submissions/2083323/bori-story

charmander: Maybe, but we have to set the limits somewhere. Notable points:

- “cover” images are already resized to a maximum of 3000 pixels tall
- WebP only supports 16383 pixels per side (which is why I picked this particular limit); AVIF’s limits are even lower

huntertur: Basing the limit on WebP's limits is reasonable to me.

PR #1472, opened 2024-11-12
Add rel="me" to profile links

huntertur:
Per discussion in the developer chatroom.

Includes a new, parametrized test for making sure `rel="me"` appears if and only if the given contact info creates a URL link.


charmander (approved):
Thanks for adding this! Testing `rel="me"` in this way is a bit odd, but I won’t turn down a new test for the otherwise untested user links 🙂

PR #1483, opened 2024-12-03
Exclude pending friends from within-friends search (#1482)

### LINKED ISSUE #1482: Searching within friends shows results for pending friend requests
If you send someone a friend request, and they haven't accepted or declined it yet, their submissions will already start to appear in search results when searching within friends. Since you and the other user aren't actually friends yet, their submissions shouldn't show up here.

Friends-only submissions are currently blanket excluded from search results (https://github.com/Weasyl/weasyl/blob/de3e03c6a4c90deffc3e10018344dd342cdbb250/weasyl/search.py#L144), so this issue can't be exploited to gain access to a user's friends-only submissions when they aren't your friend.

# Reproducing using the sample database

1. Login as `ikani`
2. Go to the Search page
    1. Set Search Terms to "ferret"
    2. Set Within to Friends
3. Verify there are no results
4. Send `fiz` a friend request
    1. Do not switch accounts to accept this; leave the request pending
6. Repeat the search with the same parameters
7. Verify results for this user now appear

huntertur:
Fixes #1482

### DIFF
--- a/weasyl/search.py
+++ b/weasyl/search.py
@@ -180,8 +180,9 @@ def _find_without_media(userid, rating, limit,
         elif within == "friend":
             # Search within friends content
             statement_from.append(
-                "INNER JOIN frienduser ON (frienduser.userid, frienduser.otherid) = (%(userid)s, content.userid)"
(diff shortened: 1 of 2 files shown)


charmander (approved):
Nice find!

PR #1485, opened 2024-12-05
Move detail tag form SCSS into its own file

huntertur:


assets/scss/site.scss
```diff
@@ -3460,7 +3332,7 @@ img + #detail-art-text {
         padding-top: 2rem;
     }
 
-    #di-info, .di-tags, #detail-manage, #detail-collection-offer, #detail-mod {
+    #di-info, #detail-manage, #detail-collection-offer, #detail-mod {
```
charmander, line 3335: I don’t think duplicating this (and the above `width: 48%` etc. rule) in a new place and separating it from its logically related elements is necessarily an improvement.

huntertur: Thank you for the feedback; I'll put those two items back in the original `site.scss` when I have time later in the day.

charmander (approved):

PR #1487, opened 2024-12-06
Add test for rejecting and undoing a suggested tag

huntertur:
While `api-unstable` implies the API could change at any time, I have already broken this feature once (in #1467, fixed by charmander in #1486), and I would like to help prevent future regressions of this feature. This test can always be deleted if it would get in the way of major API interface changes.


charmander (approved):
Thank you!

PR #1490, opened 2024-12-11
Feature/webfinger (1 / X Pull requests for Activitypub support)

Sythelux:
This is the first of probably many pull requests to add activitypub and hopefully activitystream support.

Webfinger will allow another activitypub server like mastodon to see if weasyl can "speak" activitypub it also can be used to lookup users although you will not be able to follow those, yet that will be one of the next pull requests, but all depend on this feature.

> !! we might want to add a config setting for people to opt in to federation feature.

Next steps will be:
- get user profile as json
- ability to follow a user
- outbox implementation so that everything posted on weasyl by a user can be synchronized to the fediverse
    - this is the feature that is very close to RSS feed and could be implemented along with it.

at that point weasyl would be ready to send stuff to the fediverse

inbox features would need to be discussed:
- commenting below a post would be a good post.

I have absolutely no idea how weasyl should render if someone from weasyl wants to follow someone on mastodon, but this isn't possible on wordpress either as far as I'm aware so maybe we don't need it that way.

I leave this as draft until we have discussed the feature here or on gitter.

### DIFF
--- a/weasyl/controllers/well_known.py
+++ b/weasyl/controllers/well_known.py
@@ -24,6 +24,7 @@ def webfinger(request):
     url = define.absolutify_url("")
     domain = url.replace("http://", "").replace("https://", "")
     user = profile.resolve(request.userid, userid, username)
+    # we could also support submissions/journals/collections/characters individually as actors in the future
     userprofile = profile.select_profile(user, viewer=request.userid)
     avatar_url = define.absolutify_url(userprofile['user_media']['avatar'][0]['display_url'])
     if not user:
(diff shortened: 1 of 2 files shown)


weasyl/controllers/well_known.py
```diff
@@ -0,0 +1,68 @@
...
+
+from libweasyl.exceptions import WeasylError
+from weasyl import define, profile
+
+
+def webfinger(request):
+    """Handle WebFinger requests."""
+    resource = request.params.get('resource')
+    if not resource or not resource.startswith('acct:'):
+        raise WeasylError("Invalid resource parameter")
```
huntertur, line 13: Sorry for taking so long to start to take a look at this. The string parameter that the constructor for `WeasylError` takes should match one of the keys of `weasyl.errorcode.error_messages`. The value for that key becomes the error message that is shown to the user when `WeasylError` is thrown.

PR #1493, opened 2024-12-15
Remove duplicate dl.leaders rules in site.scss

huntertur:
The removed section is a duplicate of the following section, minus one additional rule at the end:

https://github.com/Weasyl/weasyl/blob/6fd9b5969b67188e75baa09eeee0255bd36373b2/assets/scss/site.scss#L639-L669

### DIFF
--- a/assets/scss/site.scss
+++ b/assets/scss/site.scss
@@ -42,32 +42,6 @@ a:hover {
     display: none;
 }
 
-/* single, two-column dl list items with dashed borders between them */
-dl.leaders dt + dd {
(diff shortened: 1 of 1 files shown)


charmander (approved):
I looked back in the old repository, and this seems to have been the result of a merge where one branch moved the styles and added the `dl + h4, #uc-info` rule, with the commit message “Fix commission information headers” (2014-02-04). Hopefully that was just to have everything in one place and there wasn’t/isn’t some style that depends on order – everything looks right to me on the two pages that use `dl.leaders`, anyway.

Issue #1477, opened 2024-11-24
Revamp contact and social media links


huntertur:
Capturing [a discussion from 2024-11-10](https://matrix.to/#/%23Weasyl_weasyl%3Agitter.im/%24r1b8wMzhkRDBLeCSj01cduIz1MS5xVzX4L0KdLwrsfs?via=gitter.im&via=matrix.org&via=bark.lgbt) in the developer chatroom.

# Possible enhancements from discussion

- [ ] Make name/value pair orientations consistent between editor (currently vertical) and profile (currently horizontal).
- [ ] Guard against accidentally swapping names and values.
  - *"Some people add them with the site names and values swapped. The labels aren’t totally obvious enough, and it’s possible auto-recognizing URLs and splitting/moving them as they’re entered would help. Also, if the UI does auto-recognize pasted URLs and autofill the corresponding site names, how do we expose that functionality to the user? Putting the URL field first implies this flow, but then how do we support usernames? Do we need to? Predefined fields, like how FA does it, with the option to relabel and add multiples might help."* - charmander
- [ ] If a full URL to a recognized service is inserted, extract just the identifier.
  - *"The result is cleaner and more trustworthy. This also means stripping tracking parameters, mostly Twitter’s."* - charmander
- [ ] Add additional services (at least AO3, Itaku, Mastodon mentioned).
  - Bluesky added in #1439
  - Mastodon might be an issue. `@username@example.com` format is preferred to support author metadata (see below), but since anyone can be running any Fediverse software, we likely can't assume this will always map to `https://example.com/@username`. However, if we require a URL for Fediverse accounts, it might not be possible to always map back to the username-at-domain format.
- [x] Add `rel="me"` to outgoing links. (#1472)
- [ ] Add Mastodon-recognized author metadata to Weasyl embeds.
  - This seems to involve taking `<head>` and putting `<meta name="fediverse:creator" content="@username@example.com" />` in there. Assorted blog posts suggest this is a new feature added in Mastodon 4.3.

# Future goals from discussion

- [ ] Add icons to recognized services.
  - *"we do have a set of icons from weasyl3, but that needs discussion."* - charmander
- [ ] Better present contact links; they are currently small and not prominent.
- [ ] Support bidirectional verification.
  - This could potentially lead to an opt-in "last resort" option for password resets.
- [ ] Detect broken links.
- [ ] *"permalinking by user id to avoid impersonation on outdated links"*
[...]

PR #1511, opened 2024-12-27
Redesign contact links

huntertur:
Implements parts of #1477.

These changes will:
* Make editor and published link orientation consistent
* Add support for optional labels on links
* Show icons for most recognized services
  * No DeviantArt since it wants me to start a partnership with them over email before getting access to a brand kit
  * No Steam due to the following restriction in its brand rules: *"The Logo must stand alone and may not be combined with any object, including but not limited to other logos, words, graphics, photos, slogans, numbers, design features, or symbols."*
  * No YouTube because it wants me to fill out a whole form, including an employer name, to be allowed to use the logo for any purpose
  * Fur Affinity gets the low-resolution favicon instead of the current logo because I was unable to make the rainbow paw splash recognizable at low resolutions
* Support one-click removal and undoing of removals when scripting is available, supplementing the existing workflow of leaving the name or value blank on form submission to delete an entry
* Warn if a URL was likely pasted into the name field instead of the value field
* Automatically extract identifiers from recognized sites, making newly-published links look nice and tracking-free regardless of if the user typed in a username or pasted the whole profile URL
* Fix an issue in the admin tool for removing a user's contact links where an admin meaning to remove only one of the possibly-multiple links for a given site for a user ends up removing all of them upon form submission due to the form not being specific enough
* Use a combobox to show available recognized sites and their icons
  * The current `<input>` + `<datalist>` approach remains the no-scripting fallback.
  * This is my first time not only using the Web Components API, but also making a serious effort to programmatically set ARIA attributes, following the [W3C guidelines for comboboxes](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/). Because of this, it is more than possible that some of my implementation decisions here are questionable. Please let me know if I did something wrong :)

# New editor, scripting-enabled

image

# New editor, no scripting

Not pictured: `<datalist>` is still the user's friend.

image

# Published side

Notice the labels on two of the Bluesky links.

image


huntertur:
Re-tested this tonight after rebasing the branch. The one potential issue I can find is that switching to keyboard navigation can cause the page to scroll. Testing with small resolutions shows this is because the browser (at least for Firefox) wants to make the combobox be at the top of the page, if there is enough room. I'm unsure if this is an issue with my PR, or if this is expected behavior on the part of the browser. ^^;

charmander (changes requested):
Partial list of things that still need work (which I meant to get to sooner, but there’s a lot going on right now – will take care of eventually):
 
- Almost every site icon should be an SVG (Itaku and Inkbunny are acceptable exceptions), with an appropriate level of detail for the display size (Reddit is too detailed; Piczel’s SVG should be less detailed than its 300 KB PNG), and optimized. We should use SoFurry’s new logo.
- Safari doesn’t support extending built-in elements with `<input is="…">`. Have we made an explicit decision about not supporting it?
- `scrollIntoView` (regardless of options, I think?) doesn’t behave the same way as native dropdown lists.
- Table layout on narrow viewports. I don’t remember if I broke this by making it wider, but it’s important either way.
- We might want to implement a preview. It could apply sorting (highlight focused row) and display the URL to help catch mistakes. This could also be added later, though – maybe along with the future link verification feature.
- TODO: check dynamic fields’ interaction with bfcache/form state cache – is a `beforeunload` warning more important on this page?
- Is the disabled state clear enough for delete undo? We could hide the fields and replace the entire row with a `<td rowspan="3">Item deleted. <button class="link-button" type="button">Undo</button></td>`.
- We could suggest links to import from the profile text?
- Emoji/Unicode symbols in UI typically don’t look very professional. We’ll probably want to replace any remaining ones with SVG icons.

I’m going to push some changes directly here instead of opening a pull request for simplicity, but feel free to revert any.

PR #1513, opened 2024-12-27
Add embeds for characters and journals

### LINKED ISSUE #1512: Create embeds for journals and characters
Profiles and submissions currently generate embeds, but journals and characters do not.

huntertur:
Match what submissions do, as closely as possible.

This quick PR brought to you by the free Wi-Fi on a coach bus I'm currently on.

Closes #1512

### DIFF
--- a/weasyl/controllers/detail.py
+++ b/weasyl/controllers/detail.py
@@ -9,6 +9,57 @@ from weasyl.controllers.decorators import moderator_only
 from weasyl.error import WeasylError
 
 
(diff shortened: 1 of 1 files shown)


charmander (approved):
The code looks good! I wasn’t sure about using avatars for journals without further investigation – Bluesky, for example, makes a pretty terrible card with them – but now I’m not sure we can do any better, either. Other services, like Discord, create a good layout.

PR #1555, opened 2025-06-16
Add mail logging for development and clean up `weasyl.emailer`

charmander:


huntertur:
I was able to start the mail logger with `./wzl up -d mail-logger`, but I am unable to see any output from it in `./wzl logs` after changing the email address on an account. What is the correct way to view the logs for sent mail?

charmander:
It writes to `/mail`, so `./wzl exec mail-logger cat /mail/$(date -I --utc)`.

PR #1558, opened 2025-06-21
Autosize upgrades

charmander:
See individual commit messages!

### DIFF
--- a/assets/js/scripts.js
+++ b/assets/js/scripts.js
@@ -839,6 +839,8 @@ import autosize from 'autosize';
                     children.parentNode.removeChild(children);
                 }
 
+                autosize.destroy(contentField);
+
                 target.focus();
             };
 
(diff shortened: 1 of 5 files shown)


huntertur (approved):
Approved. Unrelated thought while looking at the CSS fallback comment: In older parts of `site.scss`, there are a lot of double properties in the same rule, like this:

```css
font-size: 16px;
font-size: 1rem;
```

Are these still relevant? Could they be consolidated into just either the `px` or `rem` rule? (in a separate issue / pull request, of course!)

charmander:
Yes! We haven’t needed to worry about `rem` support in a long time. I think many of them should use `px`, actually. Normally the question of units, for me, comes down to what the length is and whether it should logically scale with the text in the element – `em` if the latter *and* the number of ems can be represented as an exact floating-point number in decimal *and* the computed value is a whole number of pixels, `px` otherwise, to avoid fractional pixels[^1] and rounding errors – and I don’t usually bother with `rem`. If the length in question should logically scale with the base text size on the page and the computed value is an exact whole number, though, go for it!

[^1]: Which meant blurry lines on non-HiDPI displays, at least at one time… seems like Firefox might have fixed this? But also uneven spacing and things like that.

PR #1561, opened 2025-06-26
Improve embed posts

### LINKED ISSUE #1537: Sandbox oEmbed
- giving other services’ embeds free rein on the Weasyl origin is bad
- iframe

### LINKED ISSUE #1538: oEmbed viewer consent
Let users decide when to load other services. I’m not sure whether a list of persistently allowed services (“[ ] always load embeds from …”) would be best stored on the server or in `localStorage`, or somehow both (consider guest browsing – which might be followed by signin or signup).

I think Bluesky does this, if anyone needs UI inspiration.

charmander:
I don’t remember whether this is done, so I’ll also be reviewing it now.

> - fix Bandcamp
> - improve Bandcamp caching
> - avoid involving destination-specific DNS for Bandcamp
> - sandbox embeds
> - make loading embeds optional
> - improve error messages
> - avoid layout shift when embed loads
> - add User-Agent to outgoing embed requests
> - make embed URL parsing stricter, and use standard WHATWG URL parsing (with intent to use it for all URL parsing later on)
> 
> Preserves the existing behavior (bug) of not updating an embed’s generated thumbnail when its link is updated.

Additionally:

- improves failure behavior, exposing a fallback link instead of breaking the entire submission

resolves #1537, resolves #1538


charmander (commented):
> Additionally:
> - improves failure behavior, exposing a fallback link instead of breaking the entire submission

Okay, that’s at least one thing I hadn’t finished yet.

huntertur (approved):

PR #1566, opened 2025-07-09
Refactor site update functionality

huntertur:
Prepare for future enhancements to site updates (avoid mass `welcome` insert, make permissions more finely-grained, major/minor update distinction) with some refactoring focused on pulling everything out of `admincontrol` and `site.scss`.

There are no user-facing changes. The one admin-facing change is that site update posting is now navigated to from the site update list page instead of the admin control panel.


weasyl/controllers/routes.py
```diff
@@ -307,12 +306,16 @@
         'POST': director.directorcontrol_globaltagrestrictions_post_,
     }),
 
-    Route("/site-updates/", "site_update_list", general.site_update_list_),
+    Route("/site-updates/", "site_update_list", siteupdate.site_update_list_),
+    Route("/site-updates/new", "site_update_new", {
+        'GET': siteupdate.site_update_new_get_,
+        'POST': siteupdate.site_update_new_post_,
```
charmander, line 312: This is very minor, and it’s not like we strictly adhere to a REST-style structure at all here, but I prefer `POST /site-updates/`. Future new routes will use that structure. (`site_update_put_` was also named that way because it’s logically a PUT even though `<form>`s can’t do PUT.)

charmander (commented):
(all other changes look good!)

PR #1594, opened 2025-08-09
Fetchable metadata and accountless direct viewing for non-General posts

charmander:


huntertur:
I've run `./wzl assets` and `./wzl up -d --build web`, but I'm having trouble getting the warning to display when logged in. I'm using Fiz's account, which is set up to be General-only, and changed a submission on another account to Explicit. When I go to the URL for the submission, I can see it right away without a prompt. The prompt does show up when logged out.

huntertur (approved):

PR #1718, opened 2026-08-29
Fix prose and typos

ghost:
Cleans source comments, docstrings, tests, user-facing templates, and the two-factor authentication error identifier.

Used `codespell` to find spelling errors, `typos-cli` to scan source and identifiers, `cspell` to review repository prose, and `write-good` to review prose style. Reports from vendored JavaScript, password test data, migration hashes, and intentional project terms were excluded.


charmander (approved):
Thank you!

PR #1719, opened 2026-08-30
Convert JavaScript to TypeScript with strict mode

ghost:
Convert JavaScript to TypeScript with strict types.

Static typing helps catch errors at compilation, and improves code readability.


charmander:
Where’s this coming from? It’s a bit odd to do this out of the blue.

